PR #45, opened 2021-11-09
Docker Setup

JCSanPedro:
Create a docker image that runs a local ganache node with the contracts already deployed


paul-freeman (commented):
One little documentation change to make. There are a few other review comments that are "outdated" already - they were stuck in a pending review and I couldn't remove them.

I think it would be worth adding docker hub build hooks. It's pretty easy - you can see how it's done in the node here: https://github.com/dn3010/go-sylo-node/blob/master/hooks/build

JCSanPedro:
> I think it would be worth adding docker hub build hooks. It's pretty easy - you can see how it's done in the node here: https://github.com/dn3010/go-sylo-node/blob/master/hooks/build

Ended up configuring docker hub to trigger the image build https://hub.docker.com/repository/docker/dn3010/sylo-ethereum-testnet/builds/edit. Seems like I didn't need to add in `hooks/build` script, unless I missed something completely.

paul-freeman:
> Ended up configuring docker hub to trigger the image build https://hub.docker.com/repository/docker/dn3010/sylo-ethereum-testnet/builds/edit. Seems like I didn't need to add in `hooks/build` script, unless I missed something completely.

Maybe because it's a public repo

paul-freeman (approved):

scorpionknifes (approved):
LGTM,

Something I've noticed on https://hub.docker.com/repository/docker/dn3010/sylo-ethereum-testnet
the `latest` image seems to have an epoch duration of 80000 instead of the duration of 500 seen in `0.1.0` (might be caused by the blocktime set to auto mine every 10s)

JCSanPedro:
> the latest image seems to have an epoch duration of 80000 instead of the duration of 500 seen in 0.1.0 (might be caused by the blocktime set to auto mine every 10s)

I manually created an image with a shorter duration to make testing easier, but this shouldn't be a problem (will be adding a function to update epoch durations in a later PR)

PR #46, opened 2021-11-09
Node Listings

JCSanPedro:
This PR makes it easier to query for all nodes that have made a listing.


JCSanPedro:
Closing this as I think it's probably unnecessary and will just make the transaction of adding a listing more expensive (due to more storage costs).

For the sake of front end applications, it will be easier to just use the etherscan api to get all accounts that have made a setListing tx (and possibly just write our own backend server that periodically collects this type of information).

PR #66, opened 2022-03-09
NET-100 Fix distribution tests

JCSanPedro:
1. Up the iteration count for the distribution tests when running in circle ci. This should lower the chance that the tests will fail.
2. I also was unable to get the tests failing locally, even after running multiple times. This may be due to running node 16 locally, vs running node 14 in the circle ci image, so updated that as well.

### DIFF
--- a/test/staking.ts
+++ b/test/staking.ts
@@ -496,7 +496,9 @@ describe('Staking', () => {
     await directory.addManager(owner);
     await directory.setCurrentDirectory(epochId);
 
-    const iterations = 10000;
+    const iterations = process.env.ITERATIONS ? parseInt(process.env.ITERATIONS) : 1000;
+
+    console.log(`running all equal stake amount distribution test with ${iterations} iterations`);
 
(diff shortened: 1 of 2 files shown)


test/staking.ts
```diff
@@ -481,7 +481,7 @@ describe('Staking', () => {
       .to.be.revertedWith("Only managers of this contract can call this function");
   });
 
-  it('should distribute scan results amongst stakees proportionally - all equal [ @skip-on-coverage ]', async () => {
+  it.only('should distribute scan results amongst stakees proportionally - all equal [ @skip-on-coverage ]', async () => {
```
paul-freeman, line 484: `only` can be removed now, right?

paul-freeman (commented):
Probably need to remove `only` and make sure everything still works, but seems like a pretty simple change.

teinnt (approved):

paul-freeman (approved):

PR #67, opened 2022-03-11
NET-104 Create channel contract and update scan function

teinnt:
Summary of changes:

- Create `Channel` contract to store user's current channel on-chain.
- Update `scan(uint128 point)` to `_scan(uint256 point, uint256 epochId)`
- Create `scan(bytes memory pk, uint256 epochId, uint8 channel)` as an entry method to hash inputs and scan.
- Create `derivePoint` method to hash `public key, current epoch id, and channel` using `keccak256`.

### DIFF
--- a/contracts/Staking/Directory.sol
+++ b/contracts/Staking/Directory.sol
@@ -118,7 +118,20 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
         directories[epochId].totalStake = nextBoundary;
     }
 
-    /**
+    function scan(uint128 pk, uint8 channel)  external view returns (address stakee) {
+        uint256 point = hashPoint(pk, channel);
+        return scan(point);
+    }
+
+    function hashPoint(uint128 pk, uint8 channel) internal view returns (uint256 point) {
+        require(channel <= 48, "only channels 0-48 are supported");
+
+        bytes32 b = bytes32(keccak256(abi.encodePacked(pk, currentDirectory, channel)));
+        point = uint(b);
+        return point;
+    }
+
+     /**
      * @notice Call this to perform a stake-weighted scan to find the Node assigned
      * to the given point.
      * @dev The current implementation will perform a binary search through
@@ -126,14 +139,14 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
      * used in a transaction.
      * @param point The point, which will usually be a hash of a public key.
      */
-    function scan(uint128 point) external view returns (address stakee) {
+    function scan(uint256 point) internal view returns (address stakee) {
         if (directories[currentDirectory].entries.length == 0) {
             return address(0);
         }
 
         // Staking all the Sylo would only be 94 bits, so multiplying this with
         // a uint128 cannot overflow a uint256.
-        uint256 expectedVal = directories[currentDirectory].totalStake * uint256(point) >> 128;
+        uint256 expectedVal = directories[currentDirectory].totalStake * point >> 128;
 
         uint256 left = 0;
         uint256 right = directories[currentDirectory].entries.length - 1;
--- a/contracts/Staking/Directory.sol
+++ b/contracts/Staking/Directory.sol
@@ -128,10 +128,10 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
      * @param channel The specified channel that can effect the hash function's
      * output, which leads to different point and different node address. This will
      * allow user to switch node during the same epoch.
-     * @return stakee The address of the node that user will be joined during an epoch.
+     * @return stakee The address of the node the user will be assigned to.
      */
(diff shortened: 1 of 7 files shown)


contracts/Staking/Directory.sol
```diff
@@ -118,22 +118,35 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
         directories[epochId].totalStake = nextBoundary;
     }
 
-    /**
+    function scan(uint128 pk, uint8 channel)  external view returns (address stakee) {
+        uint256 point = hashPoint(pk, channel);
+        return scan(point);
+    }
+
+    function hashPoint(uint128 pk, uint8 channel) internal view returns (uint256 point) {
```
JCSanPedro, line 126: Better name would be `derivePoint` I think. Since the output of this function is the point value itself, and not the hash of the point.

Also `pk` should be probably just be `bytes`, and we should pass in the epochId as a parameter instead of using `currentDirectory`.

contracts/Staking/Directory.sol
```diff
@@ -118,22 +118,35 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
...
+     /**
      * @notice Call this to perform a stake-weighted scan to find the Node assigned
      * to the given point.
      * @dev The current implementation will perform a binary search through
      * the directory. This can allow gas costs to be low if this needs to be
      * used in a transaction.
      * @param point The point, which will usually be a hash of a public key.
      */
-    function scan(uint128 point) external view returns (address stakee) {
+    function scan(uint256 point) internal view returns (address stakee) {
```
JCSanPedro, line 142: I would rename this to `_scan`.

Also I think we should make this function `public`, and have it take in the `epochId` as a parameter. Then in the implementation itself, it should use the specified epochId instead of `currentDirectory` when selecting which directory to scan. This will give the ability for clients to scan using historical directories.

JCSanPedro (changes requested):
Nice. Looks like you picked things up really fast. I only suggest small changes in the API to allow clients to scan historical directories.

Also is there any tests that need to be updated? I would check that we continue to maintain 100% code coverage as well.

contracts/Staking/Directory.sol
```diff
@@ -126,27 +162,27 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
      * used in a transaction.
      * @param point The point, which will usually be a hash of a public key.
      */
-    function scan(uint128 point) external view returns (address stakee) {
-        if (directories[currentDirectory].entries.length == 0) {
+    function _scan(uint256 point, uint256 epochId) public view returns (address stakee) {
```
JCSanPedro, line 165: I don't think you want to change `uint128 point` to `uint256 point` here

contracts/Staking/Directory.sol
```diff
@@ -118,6 +118,42 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
...
+     * @notice This method is the entry point for the directory scan.
+     * It calls the derivePoint method to hash input values and send the result
+     * to the _scan method to get the associated node address.
+     * @param pk The user's public key.
+     * @param epochId The specified epochId to select which directory to scan.
+     * This will give the ability for users to scan using historical directories.
+     * @param channel The specified channel that can effect the hash function's
+     * output, which leads to different point and different node address. This will
+     * allow user to switch node during the same epoch.
+     * @return stakee The address of the node that user will be joined during an epoch.
```
JCSanPedro, line 131:
```suggestion
     * @return stakee The address of the node the user will be assigned to
 ```

File: contracts/Staking/Directory.sol (hunk 3 above)
Comment (JCSanPedro) [reply contracts/Staking/Directory.sol:165]: I think the idea of resting the point value to `uint128` is to prevent overflow when multiplying it with the totalStake of the directory

File: test/staking.ts [hunk 5]
```diff
@@ -623,7 +652,7 @@ describe('Staking', () => {
       hash.update(kp.publicKey);
       hash.update(Buffer.from([0])); // append epoch
       const point = BigNumber.from(hash.digest().subarray(0, 16));
-      const address = await directory.scan(point);
+      const address = await directory._scan(point, epochId);
       updatePoint(address);
       i++;
     }
```
JCSanPedro, line 658: Do you think you can add a test for scanning historical epochs?
So basically perform a scan at epoch 1, then initialize the next epoch, then perform a scan for epoch 1 again and check if you get the same results.

JCSanPedro:
https://961-247604013-gh.circle-artifacts.com/0/~/sylo-ethereum-contracts/coverage/index.html we lost 100% branch coverage for a couple of the files. Can you add tests to make sure we hit it?

On that note, we should probably make circle ci fail if we lose full coverage.

contracts/Staking/Directory.sol
teinnt: Sure

JCSanPedro (approved):
LGTM

PR #69, opened 2022-03-17
NET-104 Check full coverage after creating coverage report

teinnt:
Implement that the CI will fail if the coverage is not 100%.


.solcover.js
```diff
@@ -1,7 +1,35 @@
...
+      for (const [_, val] of Object.entries(valObj)) {
+        if (Array.isArray(val)) {
+          val.map(v => v === 0 && coverageFailed())
+        } else if (val === 0) {
+          coverageFailed()
+        }
+      }
+    }
+  }
+}
```
JCSanPedro, line 21: It's cool that you figured all this out, but after the coverage test is run, you can just run
`npx istanbul check-coverage --statements 100 --branches 100 --functions 100 --lines 100` (`istanbul` will need to be added as a dev dependency).

That command will check the generated coverage files.

JCSanPedro: Actually instead of `istanbul`, you can run `npx sc-istanbul ...`. `sc-istanbul` is the fork of the `istanbul` project that the `solidity-coverage` tool uses, and already exists in node_modules, so no need to add another dependency.

PR #71, opened 2022-04-28
Standardize Eslint and Prettier

JCSanPedro:
Apply eslint/prettier standards that we have in other sylo projects, such as https://github.com/dn3010/glorious-mobile.

This PR's changes doesn't need much careful review. Basically just copied the same config files from other projects, then ran

`eslint */*.ts --fix` and
`prettier */*.ts --write`


JoshD641 (approved):

teinnt (commented):
Yes, the PR looks right to me. Can you also add this file https://github.com/dn3010/glorious-mobile/blob/main/.prettierrc.js and run the `npx prettier '{*/,}*.{js,ts,json,*.js,*.ts}' --write && npx eslint '*/*.{js,ts,json,*.js,*.ts}' --fix` again?

.eslintignore
```diff
@@ -2,3 +2,7 @@ node_modules
 artifacts
 cache
 coverage
+typechain
+abi
+scripts
+*/**/*.json
```
teinnt, line 8: Can you add the last line here?

teinnt (commented):
Just one more very small thing!

.eslintignore
JCSanPedro: Good catch!

teinnt (approved):

JoshD641 (approved):

PR #72, opened 2022-04-28
NET-135 Add rewards to stake on `addStake`

JCSanPedro:
## Summary of changes

This PR consists of a UX change where when calling `addStake`, outstanding rewards are automatically added to the user's existing stake entry. This is opposed to the reward being transferred directly to the user's account.

The flow for unlocking of stake remains the same.

## Related issues and pull requests

Dependent on https://github.com/dn3010/sylo-ethereum-contracts/pull/71

## Checklist:

If not applicable to this PR, the item can be checked, crossed out, or removed.

### Documentation

- [x] Updated the readme documents (`README.md`, etc.)
- [x] Updated the spec if applicable (`docs/spec.md`, etc.)
- [x] Updated any doc comments, and checked for spelling mistakes.

### Local Build

- [x] Removed references to any local packages in build files
- [x] Ran `npm install` and no changes were made to `package-lock.json`
- [x] Ran `npm test` with no failures
- [x] Ran `npm coverage` and validated coverage has not decreased

### GitHub

- [ ] Base branch has been set correctly
- [x] PR has been assigned an owner
- [x] There are no conflicts with the base branch
- [x] Any dependent changes have been merged and published in upstream modules
- [x] Performed a self-review of my code
- [x] All commits are signed
- [ ] Additional reviews have been performed
- [x] Removed the WIP message from the top of this PR

### CircleCI

- [x] Tests are passing
- [x] Test coverage report has been reviewed


teinnt (commented):
The logic looks right to me! Do we need to make the [test coverage](https://output.circle-artifacts.com/output/job/b452d670-0703-4f3c-9544-02a46874182f/artifacts/0/~/sylo-ethereum-contracts/coverage/index.html) as 100%?

JoshD641:
The code looks good to me =)

We will need to make sure that the outstanding rewards are accounted for precisely in any client that calls addStake(). Stakers will try to "fill up" the stakee's "minimumStakeProportion" capacity exactly, so in that situation, any unintentional rounding up of the input amount will cause the transaction to fail (because of the require on line 177).

I thought about whether we should try and handle this situation in the contract code, by only adding as much stake as is allowed, but i don't think it is a good idea. The problem is that the contract can't easily tell the difference between a badly rounded input, and a situation where another staker has changed their stake in the same block as the addStake() call, where we don't know if the user would want to continue with their transaction if they had this new information.

UX wise, it's probably better to fail outright than to try and address that with code, and potentially do something the user didn't want.

JoshD641 (dismissed):

JCSanPedro:
> The logic looks right to me! Do we need to make the [test coverage](https://output.circle-artifacts.com/output/job/b452d670-0703-4f3c-9544-02a46874182f/artifacts/0/~/sylo-ethereum-contracts/coverage/index.html) as 100%?

@teinnt  The changes shouldn't have affected the test coverage, but we can look to bump it up to 100% in a different PR.

contracts/Staking/Manager.sol
```diff
@@ -149,25 +150,27 @@ contract StakingManager is Initializable, OwnableUpgradeable {
...
         Stake storage stake = stakes[stakee];
 
         uint256 currentStake = getCurrentStakerAmount(stakee, msg.sender);
 
         // automatically claim any outstanding rewards generated by their existing stake
-        _rewardsManager.claimStakingRewardsAsManager(stakee, msg.sender);
+        uint256 reward = _rewardsManager.claimStakingRewardsAsManager(stakee, msg.sender, address(this));
+
+        uint256 totalToStake = amount + reward;
+        require(totalToStake != 0, "Cannot add 0 to stake");
```
teinnt, line 162: Should we change the error message to be clearer for user?

```suggestion
        require(totalToStake != 0, "Stake amount or reward must be more than 0");
```

test/staking.ts
```diff
@@ -183,7 +183,7 @@ describe('Staking', () => {
 
   it('should not be able to add 0 stake', async () => {
     await expect(stakingManager.addStake(0, owner)).to.be.revertedWith(
-      'Cannot stake nothing',
+      'Cannot add 0 to stake',
```
teinnt, line 186: Just an update from the previous comment.

```suggestion
      'Stake amount or reward must be more than 0',
```

test/payments/ticketing.ts
```diff
@@ -891,12 +891,12 @@ describe("Ticketing", () => {
...
 
   it("should have rewards be automatically claimed when stake is updated", async () => {
-    await token.transfer(await accounts[2].getAddress(), 1000);
+    await token.transfer(await accounts[2].getAddress(), toSOLOs(1000));
     await token
       .connect(accounts[2])
       .approve(stakingManager.address, toSOLOs(1000));
 
-    await stakingManager.addStake(toSOLOs(3), owner);
+    await stakingManager.addStake(toSOLOs(10000), owner);
```
teinnt, line 899: Do you have any reason for changing it to `10000`? Does this relate to the claiming reward, or `10000` is just a better number than `3`?

contracts/Staking/Manager.sol
```diff
@@ -149,25 +150,27 @@ contract StakingManager is Initializable, OwnableUpgradeable {
 
     function addStake_(uint256 amount, address stakee) internal {
         require(stakee != address(0), "Address is null");
-        require(amount != 0, "Cannot stake nothing");
 
         Stake storage stake = stakes[stakee];
 
         uint256 currentStake = getCurrentStakerAmount(stakee, msg.sender);
```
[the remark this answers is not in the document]
JCSanPedro: > If msg.sender does not exist, it will return zero right?
Correct.

>  May I ask is there any reason for declaring stake and currentStake before totalStake?
Hmm, I don't think for any particular reason. I would probably agree that it is easier to understand if I declare after `stake` and `currentStake`

contracts/Staking/Manager.sol
JCSanPedro: Yes, good point.

test/payments/ticketing.ts
JCSanPedro: Yeah, I had to increase the value from `3` in order to meet the minimum stake requirement for the Node (it wasn't being met since the delegated staker's stake was being increased from this change). It doesn't need to be `10000` exactly, but just gave it a sufficiently high value.

teinnt: Sure, I understand now. Thank you John.

teinnt (approved):

JoshD641 (approved):
I've read through the comments, LGTM

PR #73, opened 2022-04-28
NET-140 Claim Calculation Modification

JCSanPedro:
## Motivation

The current UX for claiming rewards or changing stake is cumbersome. It's possible for a user's staking rewards to be "lost" if the user claims rewards before the end of an epoch, and tickets are continued to be redeemed after the fact. It was decided that an improved UX flow would have the user only claiming for epochs up to the previous epoch, explicitly requiring the user to wait for the current epoch to end in order to claim rewards for that epoch. This prevents any possible rewards from ever being lost to the user.

## Summary of changes

- Modify `calculateStakerClaim` to only perform the reward calculation up to the previous epoch.
- Add additional book keeping when a user makes a claim, which tracks the user's current stake as of the time making the claim. This book keeping is needed as the cumulative reward factor (CRF) cannot be used to calculate the reward for the epoch they made the claim in, as the next time they claim, their stake would have changed, and the CRF calculation assumes that their stake was static for the claim period. Thus we need to track that additional value in order to calculate their claim manually.
- Modify `calculateStakerClaim` to also manually calculate the claim for the first epoch.
- Add tests to address several edge cases.

## Related issues and pull requests

Dependent on https://github.com/dn3010/sylo-ethereum-contracts/pull/72.

## Checklist:

If not applicable to this PR, the item can be checked, crossed out, or removed.

### Documentation

- [x] Updated the readme documents (`README.md`, etc.)
- [x] Updated the spec if applicable (`docs/spec.md`, etc.)
- [x] Updated any doc comments, and checked for spelling mistakes.

### Local Build

- [x] Removed references to any local packages in build files
- [x] Ran `npm install` and no changes were made to `package-lock.json`
- [x] Ran `npm test` with no failures
- [x] Ran `npm coverage` and validated coverage has not decreased

### GitHub

- [x] Base branch has been set correctly
- [x] PR has been assigned an owner
- [x] There are no conflicts with the base branch
- [x] Any dependent changes have been merged and published in upstream modules
- [x] Performed a self-review of my code
- [x] All commits are signed
- [x] Additional reviews have been performed
- [x] Removed the WIP message from the top of this PR

### CircleCI

- [x] Tests are passing
- [x] Test coverage report has been reviewed


JoshD641:
This looks good to me. The approach is good, and it seems that the edge cases around the first claim etc are handled correctly, though it's pretty fiddly. Definitely an area for more test coverage if we have the time.

It does have a small issue with numerical precision that it inherits from the adbk fixed point division (which doesn't explicitly handle the remainder if the two numbers don't divide, it rounds down instead), but that's at the ~19th decimal place so it should have little practical impact. 

(I believe that multiplying the stake by the final CRF _before_ dividing by the initial CRF, rather than afterwards, would improve this - and our maximum possible stake values are well below the maximum that 64.64 bit fixed point can represent, so there's no risk of overflow in either case)

contracts/StateOracleDemo.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: UNLICENSED
```
JCSanPedro, line 1: Woops this file shouldn't be here.

test/payments/ticketing.ts
```diff
@@ -1506,36 +1570,227 @@ describe('Ticketing', () => {
...
+    // to claim for the first epoch
+    await epochsManager.joinNextEpoch();
+    await testClaims();
+
+    // confirm the second epoch can be claimed for
+    await epochsManager.initializeEpoch();
+    await testClaims();
+  });
+
+  it('can claim staking rewards if node already joined next epoch but skipped an epoch', async () => {
```
teinnt, line 1676:
```suggestion
  it('can claim staking rewards if node already joined next epoch but skipped the current epoch', async () => {
```

test/payments/ticketing.ts
```diff
@@ -1506,36 +1570,227 @@ describe('Ticketing', () => {
...
+          await createWinningTicket(alice, owner);
+
+        await ticketing.redeem(ticket, senderRand, redeemerRand, signature);
+      }
+    }
+
+    const testClaims = async () => {
+      const expectedStakerTwoBalance = await token
+        .balanceOf(await accounts[2].getAddress())
+        .then(b => b.add(toSOLOs(0.25 * 2500)));
```
teinnt, line 1640: Should we calculate this number `0.25`, `0.4`, and `0.35`, so that if you change the amount in the future, the test still works?

test/payments/ticketing.ts
```diff
@@ -1506,36 +1570,227 @@ describe('Ticketing', () => {
...
+      );
+
+      const expectedStakerFourBalance = await token
+        .balanceOf(await accounts[4].getAddress())
+        .then(b => b.add(toSOLOs(0.35 * 5000)));
+      await rewardsManager.connect(accounts[4]).claimStakingRewards(owner);
+      compareExpectedBalance(
+        expectedStakerFourBalance,
+        await token.balanceOf(await accounts[4].getAddress()),
+      );
```
teinnt, line 1734: This is not required because this function is only for testing, but should we reduce the repeat of checking staker's balance logic (also for the other `testClaims` function)?

test/payments/ticketing.ts
```diff
@@ -1506,36 +1570,227 @@ describe('Ticketing', () => {
...
+          await createWinningTicket(alice, owner);
+
+        await ticketing.redeem(ticket, senderRand, redeemerRand, signature);
+      }
+    }
+
+    const testClaims = async () => {
+      const expectedStakerTwoBalance = await token
+        .balanceOf(await accounts[2].getAddress())
+        .then(b => b.add(toSOLOs(0.25 * 5000)));
```
teinnt, line 1711: Should we calculate this number `0.25`, `0.4`, and `0.35`, so that if you change the amount in the future, the test still works?

teinnt (commented):
The code is very awesome! Just some small fixes, then we are good to go.

teinnt (commented):
Perfect! Only one more change and we are done!

teinnt:
Also, I am wondering if we should provide more details (or formula) for the `calculateStakerClaim` method, so that people can get an idea of how the reward claim is calculated without reading the implementation. Maybe a brief description?

JCSanPedro:
> Also, I am wondering if we should provide more details (or formula) for the calculateStakerClaim method, so that people can get an idea of how the reward claim is calculated without reading the implementation. Maybe a brief description?

The description is in the documentation here https://github.com/dn3010/sylo-ethereum-contracts/blob/master/docs/spec.md#reward-calculation-and-cumulative-reward-factor

test/payments/ticketing.ts
```diff
@@ -1125,6 +1153,72 @@ describe('Ticketing', () => {
...
+        .connect(accounts[i])
+        .approve(stakingManager.address, toSOLOs(1000));
+    }
+
+    await stakingManager.addStake(toSOLOs(1000), owner);
+    await listings.setListing('0.0.0.0/0', 1);
+
+    const accountTwoStakeProportion = 250 / 1000;
+    const accountThreeStakeProportion = 400 / 1000;
+    const accountFourStakeProportion = 350 / 1000;
```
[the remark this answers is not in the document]
JCSanPedro: I refactored and added another helper function to make all of this cleaner.

teinnt:
> > Also, I am wondering if we should provide more details (or formula) for the calculateStakerClaim method, so that people can get an idea of how the reward claim is calculated without reading the implementation. Maybe a brief description?
> 
> The description is in the documentation here https://github.com/dn3010/sylo-ethereum-contracts/blob/master/docs/spec.md#reward-calculation-and-cumulative-reward-factor

Oh right, you mean if they ask, we will give them that link. Yes, lgtm.

teinnt (approved):
I will save this PR in a folder with the description: "When John becomes serious".

JoshD641 (approved):

PR #75, opened 2022-06-28
Net 139 additional events

JCSanPedro:
### Motivation

Adding more events will allow us to have better stats/monitoring of the network, and also enables additional features for user facing applications.

### Changes
1. Update compiler version. Needed to be able to work with certain cennznet evm features.
2. Add events for when Nodes join epochs and redeem tickets.

### DIFF
--- a/contracts/Epochs/Manager.sol
+++ b/contracts/Epochs/Manager.sol
@@ -35,7 +35,7 @@ contract EpochsManager is Initializable, OwnableUpgradeable {
         uint16 decayRate;
     }
 
-    event Join (
+    event EpochJoined (
         uint256 epochId,
         address node
     );
@@ -153,7 +153,7 @@ contract EpochsManager is Initializable, OwnableUpgradeable {
     function joinNextEpoch() external {
         _directory._rewardsManager().initializeNextRewardPool(msg.sender);
         _directory.joinNextDirectory(msg.sender);
-        emit Join(currentIteration + 1, msg.sender);
+        emit EpochJoined(currentIteration + 1, msg.sender);
     }
 
     /**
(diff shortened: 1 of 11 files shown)


contracts/Epochs/Manager.sol
```diff
@@ -35,6 +35,11 @@ contract EpochsManager is Initializable, OwnableUpgradeable {
         uint16 decayRate;
     }
 
+    event Join (
```
teinnt, line 38: Should we name it `JoinNextEpoch`? I think it will be clearer for users, they will not ask why their node has not yet joined the current epoch after the transaction.

teinnt (commented):
I am thinking if we can change the event name `Join` to `JoinNextEpoch`? Otherwise the changes look good to me.

teinnt (dismissed):

contracts/Payments/Ticketing.sol
```diff
@@ -239,24 +247,38 @@ contract SyloTicketing is Initializable, OwnableUpgradeable {
...
-            incrementRewardPool(ticket.redeemer, deposit, deposit.escrow);
+            amount = deposit.escrow;
+            incrementRewardPool(ticket.redeemer, deposit, amount);
             _token.transfer(address(0x000000000000000000000000000000000000dEaD), deposit.penalty);
 
             deposit.escrow = 0;
             deposit.penalty = 0;
         } else {
-            incrementRewardPool(ticket.redeemer, deposit, epoch.faceValue);
+            amount = epoch.faceValue;
```
JoshD641, line 277: This else block is not reducing the senders deposit.escrow by the face value, while in the if block that reduction is done explicitly. I think this is a bug?

JoshD641 (changes requested):

JoshD641 (approved):
Looks good

teinnt (approved):

PR #76, opened 2022-06-28
NET-141 Public Endpoint

JCSanPedro:
### Motivation

Instead of nodes posting a multiaddr to connect to with a libp2p client, nodes will instead post a public http endpoint. This endpoint can then be queried for additional information (such as the node's multiaddr).

### Changes

Rename `multiaddr` to`publicEndpoint` in listings struct.


teinnt (approved):
Look good to me

teinnt:
It seems like there is a test failed, can you also check this John?

teinnt (approved):
LGTM

JoshD641 (approved):

PR #81, opened 2022-08-05
NET-238 Improve test coverage

teinnt:
At some point 100% coverage was lost in the ethereum contracts repo.

Add tests to ensure 100% is reached.

### DIFF
--- a/contracts/MockOracle.sol
+++ b/contracts/MockOracle.sol
@@ -87,3 +87,27 @@ contract MockOracle {
         owners[tokenId] = owner;
     }
 }
+
+contract MockOracleFail {
+    function remoteCall(
+        address target,
+        bytes memory input,
+        bytes4 callbackSignature,
+        uint256 callbackGasLimit,
+        uint256 bounty
+    ) public pure returns (uint256) {
+        require(false, "MockOracleFail always fails");
+    }
+
+    function remoteCallWithFeeSwap(
+        address target,
+        bytes memory input,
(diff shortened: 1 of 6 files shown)


contracts/MockOracle.sol
```diff
@@ -87,3 +87,27 @@ contract MockOracle {
...
+        bytes memory input,
+        bytes4 callbackSignature,
+        uint256 callbackGasLimit,
+        uint256 bounty,
+        address token,
+        uint256 maxFee
+    ) public pure returns (uint256) {
+        require(false, "MockOracleFail always fails with fee swap");
+    }
+}
```
JCSanPedro, line 113: I don't think we need to implement this MockFail contract to this line https://github.com/dn3010/sylo-ethereum-contracts/blob/master/contracts/Seekers.sol#L135 to fail.

If you just call `setOracle` on the Seekers contract, and set the `oracle` address to `addressZero`, that remote call will fail

JCSanPedro (changes requested):

contracts/MockOracle.sol
teinnt: I tried to set the mockOracle to ZeroAddress, and I have another error rather than `Oracle request failed`:
`AssertionError: Expected transaction to be reverted with Oracle request failed, but other exception was thrown: Error: Transaction reverted and Hardhat couldn't infer the reason. Please report this to help us improve Hardhat.`

I guess this is because the contract or method does not exist.

PR #86, opened 2022-08-22
NET-249 Add scan with epoch id function in Directory

teinnt:
**NOTE**: The branch ID is wrong - should be **NET-249** instead of **NET-246**.

According to https://dn3010.atlassian.net/browse/NET-146, we need another scan function with epoch ID parameter to scan a specific directory. This task is to implement something like `scanWithEpochId(uint128 point, uint256 epochId)` function that calls the scan function.

### DIFF
--- a/contracts/Staking/Directory.sol
+++ b/contracts/Staking/Directory.sol
@@ -118,35 +118,57 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
 
     /**
      * @notice Call this to perform a stake-weighted scan to find the Node assigned
-     * to the given point.
+     * to the given point of the current directory.
+     * @param point The point, which will usually be a hash of a public key.
+     */
+    function scan(uint128 point) external view returns (address stakee) {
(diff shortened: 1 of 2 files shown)


contracts/Staking/Directory.sol
```diff
@@ -118,35 +118,63 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
...
      */
     function scan(uint128 point) external view returns (address stakee) {
-        if (directories[currentDirectory].entries.length == 0) {
+        return _scan(point, currentDirectory);
+    }
+
+    /**
+     * @notice Call this to perform a stake-weighted scan to find the Node assigned
+     * to the given point of the requested directory.
+     * @dev The current implementation will perform a binary search through
```
JCSanPedro, line 134: The `@dev` part of this comment already exists for `_scan`, so no need to duplicate

JCSanPedro (approved):

PR #87, opened 2022-08-24
NET-257 Return seeker owner struct instead of owner address

teinnt:
Here is link to read the issue: https://github.com/dn3010/go-sylo-node/pull/243#discussion_r953171175

TODO:

- Deploy new contracts

Changes:
- Return `Seekers.Onwer` struct instead of `address` when calling `ownerOf`.

### DIFF
--- a/contracts/Seekers.sol
+++ b/contracts/Seekers.sol
@@ -156,19 +156,19 @@ contract Seekers is Initializable, OwnableUpgradeable {
         delete requests[requestId];
     }
 
-    function ownerOf(uint256 seekerId) external view returns (address) {
+    function ownerOf(uint256 seekerId) external view returns (Owner memory) {
         Owner memory owner = owners[seekerId];
 
         if (owner.owner == address(0)) {
-            return address(0);
+            return Owner(address(0), 0);
         }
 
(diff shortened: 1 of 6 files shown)


contracts/Seekers.sol
```diff
@@ -156,19 +156,19 @@ contract Seekers is Initializable, OwnableUpgradeable {
         delete requests[requestId];
     }
 
-    function ownerOf(uint256 seekerId) external view returns (address) {
+    function ownerOf(uint256 seekerId) external view returns (Owner memory) {
         Owner memory owner = owners[seekerId];
 
         if (owner.owner == address(0)) {
-            return address(0);
+            return Owner(address(0), 0);
```
JCSanPedro, line 163: I think we don't need this `if (owner.owner == address(0))` check anymore. We should return the `owner` value as long as it hasn't expired.

JCSanPedro (changes requested):

contracts/Seekers.sol
teinnt: Sure

JCSanPedro (approved):

PR #89, opened 2022-09-13
NET-269 Disable auto claim reward when add or unlock stake

teinnt:
Previously, we have auto claim when user adds or unlocks stake, to reduce the amount of transaction and increase stake for the node. However, this may confuse new user when they see their reward disappears. So, we will disable the auto-claim when user adding or unlocking stake.


JoshD641:
This looks good to me, but it'd be good to have John look at it as well, the CRF stuff can be pretty subtle.

PR #92, opened 2022-09-19
Net 276 seeker check

JCSanPedro:
This PR updates seeker ownership validation to just read from a standard ERC721 contract. This change can be done as we will now expect users to bridge their Seeker from Mainnet to Root.


contracts/Registries.sol
```diff
@@ -166,6 +177,51 @@ contract Registries is Initializable, OwnableUpgradeable {
...
+     * @param start The start index which is inclusive.
+     * @param end The end index which is exclusive.
+     * @return An array of Registries.
+     */
+    function getRegistries(uint256 start, uint256 end)
+        external
+        view
+        returns (address[] memory, Registry[] memory)
+    {
+        require(end > start, "end index must be greater than start index");
```
teinnt, line 200: No a review but I found this when implementing my feature. Do we allow getting exactly one registry? I.e. start = 1 and end = 1?

contracts/Registries.sol
```diff
@@ -166,6 +177,51 @@ contract Registries is Initializable, OwnableUpgradeable {
...
+        }
+
+        return (_nodes, _registries);
+    }
+
+    /**
+     * @notice Returns the total number of registered nodes.
+     * @return The number of registered nodes.
+     */
+    function getTotalNodes() external view returns (uint256) {
```
teinnt, line 221: Should we name it to `getNodesCount()`?

contracts/TestSeekers.sol
```diff
@@ -0,0 +1,19 @@
...
+  constructor() ERC721("Seekers", "SEEKERS") {}
+
+  function mint(address to, uint256 tokenId) external {
+    _safeMint(to, tokenId);
+  }
+
+  function exists(uint256 tokenId) external view returns (bool) {
+    return _exists(tokenId);
+  }
+}
```
teinnt, line 19: Maybe we need to add one more line at the end.

contracts/Registries.sol
teinnt, line 200: I think we should use uppercase to make it consistent with other error messages.

```suggestion
        require(end > start, "End index must be greater than start index");
```

contracts/Registries.sol
```diff
@@ -166,6 +177,51 @@ contract Registries is Initializable, OwnableUpgradeable {
...
+     */
+    function getRegistries(uint256 start, uint256 end)
+        external
+        view
+        returns (address[] memory, Registry[] memory)
+    {
+        require(end > start, "end index must be greater than start index");
+        require(
+            end <= nodes.length,
+            "end index cannot be greater than total number of registered nodes"
```
teinnt, line 203:
```suggestion
            "End index cannot be greater than total number of registered nodes"
```

contracts/Registries.sol
teinnt, line 221: I think we can name it to `getNodesCount()`

test/registries.ts
```diff
@@ -68,6 +66,89 @@ describe('Registries', () => {
...
+      );
+      assert.equal(
+        result[1][i - 5].publicEndpoint,
+        `http://api/${i}`,
+        'Expected correct registry to be returned',
+      );
+    }
+
+    await expect(registries.getRegistries(8, 5)).to.be.revertedWith(
+      'end index must be greater than start index',
```
teinnt, line 144:
```suggestion
      'End index must be greater than start index',
```

test/registries.ts
```diff
@@ -68,6 +66,89 @@ describe('Registries', () => {
...
+    );
+
+    await expect(registries.getRegistries(8, 21)).to.be.revertedWith(
+      'end index cannot be greater than total number of registered nodes',
```
teinnt, line 148:
```suggestion
      'End index cannot be greater than total number of registered nodes',
```

teinnt (commented):
Just a little more to change, otherwise looking good to me.

contracts/Epochs/Manager.sol
```diff
@@ -39,7 +39,7 @@ contract EpochsManager is Initializable, OwnableUpgradeable {
 
     Registries public _registries;
 
-    Seekers public _seekers;
+    IERC721 public _rootSeekers;
```
JoshD641, line 42: Tiny detail, but I can see the name rootSeekers being confusing if we're ever running this on something other than root network. 

Maybe bridgedSeekers instead? That way it's still clear that they're on the same blockchain as the contracts, and that the data structure contains only the subset of the seekers that have been bridged.

contracts/Registries.sol
```diff
@@ -138,9 +149,9 @@ contract Registries is Initializable, OwnableUpgradeable {
         );
 
         // Now verify the seeker account actually owns the seeker
-        Seekers.Owner memory owner = _seekers.ownerOf(seekerId);
+        address owner = _rootSeekers.ownerOf(seekerId);
```
JoshD641, line 152: What does ownerOf(seekerID) return when the specified seeker hasn't been bridged? 

We could explicitly handle that case, separately from "sorry, the owners don't match", so that the dashboard UI can give a more helpful error message.

JoshD641 (changes requested):
Looks good to me - just a few comments about naming and more specific error messages for UX, nothing major.

contracts/Registries.sol
JCSanPedro: It is address zero, but it could potentially be transferred (burned) to that address. I guess that would be an odd edge case though

contracts/Epochs/Manager.sol
JCSanPedro: It might be just better to call it `_seekers` in this case (maybe we will be on Eth mainnet :P)

contracts/Registries.sol
JoshD641: Yea, that does seem unlikely to me. 

I think interpreting "owned by address zero" as "not bridged" is better UX wise than interpreting "not bridged" as "you don't own that seeker", though that assumption should probably be made in the dashboard rather than the contracts. If we just return an error "Owned by address 0", the contracts can be unambiguous, and the dashboard can interpret that how it wants.

JoshD641 (approved):

contracts/Registries.sol
JCSanPedro: Yeah the dashboard shouldn't. get into this case because it only shows seekers owned by the user, so it wouldn't register a seeker not owned by the user anyway.

contracts/Registries.sol
JCSanPedro: Not really a significant difference between either of these names, but just personally prefer the world `total` over `count`

teinnt (approved):
Awesome! LGTM.

PR #93, opened 2022-10-27
Net 319 Threshold behaviour update

JCSanPedro:
Currently Nodes that do not meet the minimum stake threshold can not join epochs. Additionally, the contracts prevent stakers from adding more stake if it will cause the Node to no longer meet the minimum.

This PR changes the behaviour to allow Nodes to join epochs with less than the minimum, with the caveat that their stake would be reduced proportionally.

For example, if the minimum threshold is 20% of owned stake, and the Node only owns 10% (50% of the threshold), then the Node's total stake will be reduced by 50% when joining an epoch.

This PR also includes an additional change to make it easier to query the total amount of rewards gained in a particular epoch.


contracts/Payments/Ticketing/RewardsManager.sol
```diff
@@ -80,6 +80,16 @@ contract RewardsManager is Initializable, OwnableUpgradeable, Manageable {
...
+    /**
+     * @notice Tracks total accumulated rewards in each epoch
+     */
+    mapping(uint256 => uint256) public totalEpochRewards;
+
+    /**
+     * @notice Tracks total accumulated staking rewards in each epoch
+     */
+    mapping(uint256 => uint256) public totalEpochStakingRewards;
+
```
JoshD641, line 92: Doing this in the contracts, rather than an external service watching the chain, makes us rely more on low gas prices on whatever chain we use. Should we consider having the node do this instead?

contracts/Staking/Directory.sol
```diff
@@ -94,11 +94,24 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
...
+        } else {
+            // if the node does meet the minimum stake proportion, then we reduce
+            // the stake used to join the epoch proportionally
+            uint256 currentStake = _stakingManager.getCurrentStakerAmount(stakee, stakee);
+            uint16 ownedStakeProportion = SyloUtils.asPerc(
+                uint128(currentStake),
+                totalStake
+            );
+
+            joiningStake = totalStake * ownedStakeProportion / _stakingManager.minimumStakeProportion();
```
JoshD641, line 112: This phrasing of the adjustment is potentially dangerous - if there is a bug in checkMinimumStakeProportion() causing it to return false incorrectly, then the ratio would be greater than 1, and this line would increase joiningStake above totalStake, giving a node more traffic than they should get.

We could add a require(joiningStake <= totalStake) to catch this, or rephrase the adjustment so that it subtracts away from totalStake.

contracts/Payments/Ticketing/RewardsManager.sol
JCSanPedro: So the way for a node to do this is to retrieve all ticket redemption events for the block period corresponding to an epoch, then adding up all the tickets. This is a pretty common operation, and would end up being a lot of data to query each time.

Since we will be on a low cost blockchain in the immediate future, I felt that adding an extra 10k to the gas limit for each redemption will be better than having to retrieve all the event logs each time.

Also we will have an external service that we control that will be watching ticket events and caching them. The dashboard will be able to query this service for metrics like the total tickets redeemed by a node, but I don't want the Nodes themselves to be relying on any external service that we control.

contracts/Staking/Directory.sol
JCSanPedro: Good point, will add in that require call. I think refactoring the calculation will be tricky without running into underflow edge cases.

teinnt (commented):
I see that we only reduce the stake amount when joining next director. So in the next epoch, if the node adds stake, the `joining stake` will be `total stake` and everything will work as normal. Yeah, I am happy with this.

contracts/Payments/Ticketing/RewardsManager.sol
JoshD641: That makes sense - this way seems to be the best approach on balance, thanks for explaining =)

JoshD641 (dismissed):
Looks good =)

contracts/Staking/Directory.sol
```diff
@@ -95,22 +95,36 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
...
+
+        uint256 joiningStake = 0;
+        if (ownedStakeProportion >= minimumStakeProportion) {
+            joiningStake = totalStake;
+        } else {
+            // if the node does meet the minimum stake proportion, then we reduce
+            // the stake used to join the epoch proportionally
+            joiningStake = totalStake * ownedStakeProportion / minimumStakeProportion;
+        }
+        require(joiningStake > 0, "Can not join directory for next epoch without any stake");
```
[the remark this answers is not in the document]
JCSanPedro: Why would we need that restriction here?

I think you can't reach this branch in the code even if `minimumStakeProportion == 0`, since the `ownedStakeProportion` will always be greater than or equal to 0 anyway.

Anyway it is unlikely, but we might want to be able to set `minimumStakeProportion` to 0 (i.e, remove that requirement).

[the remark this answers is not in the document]
teinnt: In my opinion, it is unsafe when we divide a number by 0. I tried to test it with Remix, and it does not allow compiling the line dividing by zero, so I asked you if we should have a restriction for it.

That's what I tested
Screen Shot 2022-11-30 at 1 05 09 PM

That is the error
Screen Shot 2022-11-30 at 1 05 30 PM

When I changed to something else
Screen Shot 2022-11-30 at 1 06 05 PM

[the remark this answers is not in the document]
JCSanPedro: I definitely agree on regarding division by 0, but I think it is impossible to call `joiningStake = totalStake * ownedStakeProportion / minimumStakeProportion;` where `minimumStakeProportion == 0`, because of the check above `if (ownedStakeProportion >= minimumStakeProportion) {`.

At least I don't see a case where that could happen. But if you know one, please let me know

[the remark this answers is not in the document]
teinnt: Oh yes, that makes sense. Thank you very much John!

teinnt (approved):
LGTM

PR #94, opened 2022-11-08
NET-77-Local-rootnet-docker

teinnt:
### Motivation ###

Since we are moving to Rootnet blockchain, we should have our testnet running on local Rootnet. This PR is to build and deploy contracts for the local rootnet docker image, and update the current ganache docker image to use the new contracts.

### Summary Changes ###

- Create a rootnet testnet with contracts deployed and auto mint 10 seekers for the deployer
- Update ganache testnet with new contracts deployed and auto mint 10 seekers for the deployer
- Update scripts js files to ts
- Add utils file that export `mintSeeker` function


deploy/00_deploy_phase_two.ts
```diff
@@ -114,73 +110,121 @@ async function deployPhaseTwoContracts(
...
   const directory = (await upgrades.deployProxy(DirectoryFactory, undefined, {
     initializer: false,
   })) as Directory;
+  await directory.deployed();
 
   logDeployment('Directory', directory.address);
 
-  await epochsManager.initialize(
-    seekers.address,
+  let tx = await registries.initialize(
```
JCSanPedro, line 125: Don't really like using the `let` keyword in general but I see why you used it here. Anyway, if we just need the tx variable to call `tx.wait()`, we can do this instead
```
await registries.initialize(....).then(tx => tx.wait());
```

deploy/00_deploy_phase_two.ts
```diff
@@ -114,73 +110,121 @@ async function deployPhaseTwoContracts(
...
+
+  tx = await rewardsManager.addManager(stakingManager.address);
+  await tx.wait();
+
+  tx = await rewardsManager.addManager(epochsManager.address);
+  await tx.wait();
 
   console.log('Added managers to rewards manager contract');
 
+  if (config.Seekers.shouldMint) {
```
JCSanPedro, line 223: I don't like this being part of the deployment script. I think for testnets, we can just define another javascript file to run the code to do the minting, as opposed to this conditional check.

deploy/genesis.config.ts
```diff
@@ -73,20 +72,20 @@ const GenesisParameters: ContractParameters = {
...
+    faceValue: hre.ethers.utils.parseEther('10000'),
+    baseLiveWinProb: BigNumber.from(2).pow(128).sub(1), // 100%
     expiredWinProb: BigNumber.from(2).pow(128).sub(1).div(1000),
     ticketDuration: 80000,
     decayRate: 8000,
   },
 
   Ticketing: {
-    unlockDuration: 80000,
+    unlockDuration: 30, // 150 seconds
```
JCSanPedro, line 83: Let's revert the changes to these parameters when merging into `master`. We can keep it for `deploy-contracts`, as that is the branch we use for testing, but not for `master`.

docker/rootnet-with-protocol/README.md
```diff
@@ -0,0 +1,45 @@
...
+  "rewardsManager": "0x4002F76bA7db46b3Dc79507C11A08Fa389b9d6CB",
+  "epochsManager": "0x161b5FA7CCCf8477Dc892941e303A159EC7D3e57",
+  "stakingManager": "0x41Cd8EB92BD2f7d5eAa1202666AB9EC9F7058151",
+  "seekers": "0x50f15F0a0B798b6E3A248Fccc26b6de636b7c3ef"
+}
+```
+
+The chain state is saved in the directory `/mnt/data`.
+
+The `private key` used to when starting the node and also
```
JCSanPedro, line 25:
```suggestion
The `private key` used when starting the node and also
```

scripts/contract_interaction.ts
```diff
@@ -0,0 +1,33 @@
...
+  const xrpToken = new Contract(
+    xrpTokenAddress,
+    erc20Abi,
+    deployer,
+  ) as SyloToken;
+  const balance = await xrpToken.balanceOf(deployer.address);
+  console.log('balance', deployer.address, balance.toString());
+}
+
+main();
```
JCSanPedro, line 33: What's the point of this file? I think any like utility scripts we can keep in our `sylo-ethereum-utils` repo

JCSanPedro (changes requested):
Cool, just some minors otherwise looks good.

Thank you for also updating the `scripts` files to be typescript.

scripts/contract_interaction.ts
teinnt: Oh yeah, it is for me to test contract methods. I will remove it.

JCSanPedro (approved):

PR #95, opened 2022-12-01
NET-328 Fix Reward Claim

JCSanPedro:
The previous changes to the reward calculation introduced a bug that mistakenly omitted the initial claim value from the final calculation. This PR resolves that issue and adds another test case to ensure that particular edge case does not get missed.


JoshD641 (approved):

teinnt (approved):
Awesome!

PR #97, opened 2022-12-14
NET-346 Amend behaviour in `setRegistry` to automatically revoke seekers that have already been registered

Nick95550:
## Motivation
It is possible to register the same seeker multiple times as long as it is owned by the specified seeker account.
This can lead to confusion where there exists more than one Registry with the same seeker Id. It is better to remove
this so that a single seeker can only be registered in a single registry at a time.

## Summary of changes
A new mapping in the Registries contract keeps track of the seeker account associated with the seeker ID.
Using this mapping, when a seeker is transferred from one account to another and registered the old registry has its seeker Id field reset.

To test this a new test case has been written and implemented that mints a seeker, transfers it and ensures that the old registry is successfully cleared. 

A call to revokeSeekerAccount also resets this new mapping 

## Checklist

- [X] Updated the readme documents (`README.md` etc.)
- [X] Updated `CHANGELOG.md` with new items at top of list

### Local Build

- [X] Run npm run test


deploy/hardhat_deployment_phase_two.json
```diff
@@ -0,0 +1,12 @@
...
+ "token": "0x5FbDB2315678afecb367f032d93F642f64180aa3",
+ "registries": "0xDc64a140Aa3E981100a9becA4E685f962f0cF6C9",
+ "ticketing": "0x4ed7c70F96B99c776995fB64377f0d4aB3B0e1C1",
+ "ticketingParameters": "0x0165878A594ca255338adfa4d48449f69242Eb8F",
+ "directory": "0x9A676e781A523b5d0C0e43731313A708CB607508",
+ "rewardsManager": "0xA51c1fc2f0D1a1b8494Ed1FE312d7C3a78Ed91C0",
+ "epochsManager": "0x2279B7A0a67DB372996a5FaB50D91eAA73d2eBe6",
+ "stakingManager": "0x610178dA211FEF7D417bC0e6FeD39F05609AD788",
+ "seekers": "0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512"
+}
```
JCSanPedro, line 12: No need to track this file in our repo

deploy/nikau_deployment_phase_two.json
```diff
@@ -1,12 +1,12 @@
...
+ "token": "0xcccCccCC000042B4000000000000000000000000",
+ "registries": "0x37072Cada0dB0183C7A7f032F9734e30eed0437C",
+ "ticketing": "0xc70cE1D304b224c1ac0dF68bB9E1c97f8767B939",
+ "ticketingParameters": "0xF7bbb3102C93D5Ead06d68F179494368522523e3",
+ "directory": "0x4AB2eddcc10E2D3cE4F8E3314B579A4c96095357",
+ "rewardsManager": "0x041c2BD2A37151FCD6bDAeB0be25E3FE15731d5A",
+ "epochsManager": "0xA5E4c20Ad3baAdd7404d33753f7Af9612b50175a",
+ "stakingManager": "0xD5eCb504b97bA845f22e1E95565C86fD469a6248",
+ "seekers": "0xC65fDC6c38D0a1d3524aE54ba205BDE197AbddbA"
+}
```
JCSanPedro, line 12: Again, this change is inconsequential but still unrelated.

test/registries.ts
```diff
@@ -277,4 +277,60 @@ describe('Registries', () => {
       'Public endpoint can not be empty',
     );
   });
+
+  it("Doesn't exists that more than one Registry has same seeker ID", async() => {
```
JCSanPedro, line 281:
```suggestion
  it("can't register a seeker using the same seeker ID", async() => {
```

test/registries.ts
```diff
@@ -277,4 +277,60 @@ describe('Registries', () => {
...
+    // Gets the current block number and turns it into a string
+    const block = await ethers.provider.getBlockNumber(); 
+    const blockString = block.toString();
+
+    // Gets the prefix for the proof messages from the Registries contract "This message allows your seeker to be used to operate your node."
+    const proofMessagePrefix = await registries.getPrefix();
+
+    // Constructs the proof message
+    // Signs the proof message returns a Raw Signature promise 65 byte DataHexString 
+    const proofMessage = `${proofMessagePrefix}:${tokenID}:${seekerAddressLower}:${blockString}`;
```
JCSanPedro, line 306: Can we just reuse this function when setting the seeker https://github.com/dn3010/sylo-ethereum-contracts/blob/master/test/utils.ts#L160?

test/registries.ts
```diff
@@ -277,4 +277,60 @@ describe('Registries', () => {
...
+
+    // Connects the second seeker account, seekerAccountTwo, and sets it as a seeker account
+    // retreives the registry for the second seeker account
+    await registries.connect(seekerAccountTwo).setSeekerAccount(seekerAddressTwo, tokenID, block, signatureTwo);
+
+    // Get the registries for the two seeker accounts
+    const regoSeekerAccountTwo = await registries.getRegistry(seekerAddressTwo);
+    const regoSeekerAccount = await registries.getRegistry(seekerAddress);
+
+    // Tests that the registry for both seekers accounts has the same seekerID (BAD, NEED TO FIX)!!!!!
```
JCSanPedro, line 331: Does this comment need to be updated?

test/registries.ts
```diff
@@ -277,4 +277,60 @@ describe('Registries', () => {
...
+    const seekerAddress = await seekerAccount.getAddress(); // Get the address for the account in use
+    const seekerAddressLower = seekerAddress.toLowerCase(); // Sets the seekerAddress to lower case as it is a string promise
+
+    const seekerAccountTwo = accounts[5];
+    const seekerAddressTwo = await seekerAccountTwo.getAddress();
+    const seekerAddressLowerTwo = seekerAddressTwo.toLowerCase(); // Sets the seekerAddress to lower case as it is a string promise
+
+    /* 
+    Repeating this process for my own understanding
+    */
```
JCSanPedro, line 292: Multi-line comments using `/* ... *../` are usually reserved for function documentation

contracts/Registries.sol
```diff
@@ -166,6 +176,8 @@ contract Registries is Initializable, OwnableUpgradeable {
         );
 
         registry.seekerAccount = address(0);
+        seekerRegistration[registry.seekerId] = address(0);
```
JCSanPedro, line 179: I think we can also add in here
`registry.seekerId = 0`

contracts/Registries.sol
```diff
@@ -149,12 +154,17 @@ contract Registries is Initializable, OwnableUpgradeable {
...
         address owner = _rootSeekers.ownerOf(seekerId);
-
+        
         require(seekerAccount == owner, "Seeker account must own the specified seeker");
 
         registries[msg.sender].seekerAccount = seekerAccount;
         registries[msg.sender].seekerId = seekerId;
+
+        registries[seekerRegistration[seekerId]].seekerId = 0;
+        seekerRegistration[seekerId] = seekerAccount;
```
JCSanPedro, line 166: I think this needs to be
`seekerRegistration[seekerId] = msg.sender;`.

The `seekerAccount` represent's the node runner's personal account. `msg.sender` represents the node's own account as this function will be sent from the itself.

A node runner can run multiple nodes but their seeker should be associated with only one node account (`msg.sender`) at any given time.

JCSanPedro (changes requested):

JCSanPedro:
Good job learning all the typescript/nodejs/ethers.js stuff!

JCSanPedro (changes requested):
Looks good, just a couple nitpicky comments I've left.

Also, I see in this PR you have deleted `.openzeppelin/ropsten.json` and `deploy/nikau_deploy_phase_two.json`. Since deleting those files are unrelated to the PR, we *shouldn't* include, but because we kind of need to remove them anyway, I'll allow this time around.

test/staking.ts
```diff
@@ -650,7 +652,7 @@ describe('Staking', () => {
     // process epoch 2
     const amountEpochTwo = [50, 100, 100, 300, 450];
     for (let i = 0; i < amountEpochTwo.length; i++) {
-      await addStakeAndJoinEpoch(accounts[i], amountEpochTwo[i], i);
+      await addStakeAndJoinEpoch(accounts[i], amountEpochTwo[i], i + seekerIdOffset);
```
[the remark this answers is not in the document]
Nick95550: Yea but I believe the problem is that the same seeker ID is being used in the previous loop here https://github.com/dn3010/sylo-ethereum-contracts/blob/master/test/staking.ts#L644. So this will register a different seeker to the account.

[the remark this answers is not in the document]
JCSanPedro: I don't think that problem is happening because this loop is for joining the next epoch (and no seekers are in use for the next epoch). Also, it wouldn't make sense anyway since you're only adding `1` to the iteration value, so some of the seeker ids would be reused.

I just checked out the code. It's actually a bug in the logic. If the same node registers with the same seeker Id twice, then when it hits this line in the code
`registries[seekerRegistration[seekerId]].seekerId = 0;`
It basically sets its own registries seekerId to 0.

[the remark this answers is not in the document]
Nick95550: Yea I suppose I hadn't thought that the account would set the same seeker twice in a row, so implemented this work around that would ensure that a different seeker is set to the same account during testing. I'll change the logic now using your suggestion.

JCSanPedro (approved):

test/registries.ts
```diff
@@ -277,4 +277,28 @@ describe('Registries', () => {
       'Public endpoint can not be empty',
     );
   });
+
+  it("can't register a seeker using the same seeker ID", async() => {
```
teinnt, line 281: I think we _can_ register a seeker using the same seeker ID, because there is no error in the code. Should we say 

```suggestion
  it("registered node seeker id will be reset if a new node registers with the same seeker id", async() => {
```

test/registries.ts
```diff
@@ -277,4 +277,28 @@ describe('Registries', () => {
...
+    const seekerAccount = accounts[3];
+    
+    const account = accounts[2];
+    const accountAddress = await account.getAddress();
```
teinnt, line 285: Should we name them `accountOne` and `accountAddressOne` so it is more readable and easier to compare? I think would be good to update `regoSeekerAccount` to `regoSeekerAccountOne` too.

test/registries.ts
```diff
@@ -277,4 +277,28 @@ describe('Registries', () => {
...
+    const accountAddress = await account.getAddress(); 
+
+    const accountTwo = accounts[4];
+    const accountAddressTwo = await accountTwo.getAddress();
+
+    const tokenID = 100;
+
+    await utils.setSeekerRegistry(registries, seekers, account, seekerAccount, tokenID);
+
+    await utils.setSeekerRegistry(registries, seekers, accountTwo, seekerAccount, tokenID);
```
teinnt, line 294: Hmmm, I am not sure if you use prettier, can you check if you have it installed and enabled? As this is the format we should have with prettier (it auto formatted when I pressed Command+S). Can you try to run this command `npx prettier ./test --write`?

Screen Shot 2022-12-19 at 4 33 10 PM

test/registries.ts
```diff
@@ -277,4 +277,28 @@ describe('Registries', () => {
...
+
+    const regoSeekerAccount = await registries.getRegistry(accountAddress);
+    const regoSeekerAccountTwo = await registries.getRegistry(accountAddressTwo);
+
+    // tests that the registry for both seeker accounts don't have the same seekerID
+    expect(regoSeekerAccount.seekerId).not.equal(regoSeekerAccountTwo.seekerId);
```
teinnt, line 300: Do you think we need this check, because if other checks pass (comparing with 0 and 100), then we know they are not equal right?

test/registries.ts
```diff
@@ -277,4 +277,28 @@ describe('Registries', () => {
...
+    expect(regoSeekerAccount.seekerId).to.equal(0);
+    expect(regoSeekerAccountTwo.seekerId).is.equal(100);
```
teinnt, line 302:
```suggestion
    expect(regoSeekerAccountTwo.seekerId).is.equal(tokenID);
```

contracts/Registries.sol
```diff
@@ -166,6 +175,8 @@ contract Registries is Initializable, OwnableUpgradeable {
         );
 
         registry.seekerAccount = address(0);
+        seekerRegistration[registry.seekerId] = address(0);
+        registry.seekerId = 0;
```
teinnt, line 179: Not necessary, but we can change it like this so it is more readable and consistent with the order we have in `setSeekerAccount`.

```suggestion
        registry.seekerId = 0;
        registry.seekerAccount = address(0);

        seekerRegistration[registry.seekerId] = address(0);
```

teinnt (commented):
Well done! Only some small things left and this PR is ready.

test/registries.ts
Nick95550: Yea that is a better description I think

test/registries.ts
Nick95550: This command didn't do anything for me, I think I have prettier installed though.

contracts/Registries.sol
Nick95550: I think that making this change will produce a bug in the logic, as the registry.seekerId will be set to zero before it is used in the seekerRegistration mapping

teinnt: Right, my bad. Thanks Nic!

teinnt (approved):
LGTM

JCSanPedro (approved):

PR #98, opened 2022-12-14
NET-352 remove minDelegatedStake from registries contract

Nick95550:
## Motivation
The minDelegatedStake field is intended to be configured by nodes to set a minimum Sylo amount that a delegate must provide when staking. It was decided that we did not need to use this value for phase 2, as its main purpose is to prevent the list of delegated stakers being too long, which could have an implication on gas costs, but this no longer applies.

However the field still exists in the Registry, and is still passed in when calling register. We should remove this field from the contract for phase 2 as it creates unnecessary confusion.

## Summary of changes
- Removed the minDelegatedStake field from the Registry struct within the Registries contract
- Removed the minDelegatedStake parameter within the register function call
- Amended the test cases by removing the now extra parameter

### Documentation

- [X] Updated the readme documents (`README.md` etc.)
- [X] Updated `CHANGELOG.md` with new items at top of list

### Local Build

- [X] Execute npm run test

### Jira
- [X] PR name matches Jira task `NET-XXX`
- [X] Branch name matches Jira task `NET-XXX`

### GitHub

- [X] Target branch has been set correctly
- [X] PR has been rebased onto target branch
- [X] Author has performed a self-review of the code
- [X] Removed the WIP message from the top of this PR

### DIFF
--- a/contracts/Registries.sol
+++ b/contracts/Registries.sol
@@ -36,9 +36,6 @@ contract Registries is Initializable, OwnableUpgradeable {
         // This value is currently locked to the default payout percentage
         // until epochs are implemented.
         uint16 payoutPercentage;
-        // The minimum amount of stake that is required to
-        // add delegated stake against this node
-        uint256 minDelegatedStake;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


JCSanPedro:
Sweet, looks good. I'm going to update the scope of this task (NET-352) to also include updating the go-sylo-network-client repo and it's use in the go-sylo-node repo.

Specifically this function https://github.com/dn3010/go-sylo-network-client/blob/main/client.go#L569.

So there should be 3 PRs, one for this repo, one for go-sylo-network-client, and one for the go-sylo-node repo.

JCSanPedro (approved):

teinnt (commented):
The code LGTM. Should we remove `minDelegatedStake` in `docs/spec.md` too?

teinnt (approved):
Awesome.

JCSanPedro:
Good job updating the spec docs

PR #100, opened 2023-01-18
Net 396 update signing message

Nick95550:
## Todo
I expect that some renaming will need to be done and or removing of the added test case within the registries typescript file

- [X] add some items to the todo list

## Motivation
The current message shown to the user is boring, and also does not do a good enough job of explaining the purpose of signing the message. The marketing has given us an updated message to use which is not written into the Registries contract.

## Summary of changes
- Removed old seeker ownership prefix with new variable to hold the seeker ownership message.
- Created function to update ownership message whenever the seeker registry is set, important for the block number
-  Wrote a test case to check the block number would be correct and display the message

### Local Build

- [X] Ran the test suite locally (`npm test`)

### Jira
- [X] PR name matches Jira task `NET-XXX`
- [X] Branch name matches Jira task `NET-XXX`

### GitHub

- [X] Target branch has been set correctly
- [X] PR has been rebased onto target branch
- [X] Author has performed a self-review of the code
- [X] Removed the WIP message from the top of this PR

### DIFF
--- a/contracts/Registries.sol
+++ b/contracts/Registries.sol
@@ -16,8 +16,8 @@ import "./ECDSA.sol";
 contract Registries is Initializable, OwnableUpgradeable {
     using ECDSA for bytes32;
 
-    string public constant SEEKER_OWNERSHIP_PREFIX =
-        "This message allows your seeker to be used to operate your node.";
+    // string public constant SEEKER_OWNERSHIP_PREFIX =
+    //     "This message allows your seeker to be used to operate your node.";
 
     struct Registry {
         // Public http/s endpoint to retrieve additional metadata
@@ -37,6 +37,13 @@ contract Registries is Initializable, OwnableUpgradeable {
         // until epochs are implemented.
         uint16 payoutPercentage;
     }
+
+    /**
+     * @notice Holds the seeker ownership message that includes
+     * the users seeker_id, node address, and block the message
+     * is to signed on
+     */
+    string private seeker_ownership_message;
 
     /**
      * @notice ERC721 contract for bridged Seekers. Used for verifying ownership
@@ -158,6 +165,8 @@ contract Registries is Initializable, OwnableUpgradeable {
         registries[msg.sender].seekerId = seekerId;
 
         seekerRegistration[seekerId] = msg.sender;
+
+        updatePrefix();
     }
 
     function revokeSeekerAccount(address node) external {
@@ -227,11 +236,29 @@ contract Registries is Initializable, OwnableUpgradeable {
         return nodes.length;
     }
 
+    function updatePrefix() private {
+        string
+            memory prefixLineOne = unicode"🤖 Hi frend! 🤖\n\n📜 Signing this message proves that you're the owner of this Seeker NFT and allows your Seeker to be used to operate your Seeker's Node. It's a simple but important step to ensure smooth operation.\n\nThis request will not trigger a blockchain transaction or cost any gas fees.\n\n🔥 Your node's address: ";
+        string memory prefixLineTwo = unicode"\n\n🆔 Your seeker id: ";
+        string memory prefixLineThree = unicode"\n\n📦 The block this message was signed: ";
+
+        seeker_ownership_message = string(
+            abi.encodePacked(
+                prefixLineOne,
+                Strings.toHexString(msg.sender),
+                prefixLineTwo,
+                Strings.toString(registries[msg.sender].seekerId),
+                prefixLineThree,
+                Strings.toString(block.number)
+            )
+        );
+    }
+
     /**
      * @notice Retrieves the prefix used for creating proofs.
      */
-    function getPrefix() public pure returns (string memory) {
-        return SEEKER_OWNERSHIP_PREFIX;
+    function getPrefix() public view returns (string memory) {
+        return seeker_ownership_message;
     }
 
     /**
@@ -246,10 +273,10 @@ contract Registries is Initializable, OwnableUpgradeable {
         uint256 seekerId,
         address node,
         uint256 proofBlock
-    ) public pure returns (bytes memory) {
+    ) public view returns (bytes memory) {
         return
             abi.encodePacked(
-                SEEKER_OWNERSHIP_PREFIX,
+                seeker_ownership_message,
                 ":",
                 Strings.toString(seekerId),
                 ":",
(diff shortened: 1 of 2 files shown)


contracts/Registries.sol
```diff
@@ -227,11 +233,29 @@ contract Registries is Initializable, OwnableUpgradeable {
         return nodes.length;
     }
 
+    function updatePrefix() private {
+        string
+            memory prefixLineOne = unicode"🤖 Hi frend! 🤖\n\n📜 Signing this message proves that you're the owner of this Seeker NFT and allows your Seeker to be used to operate your Seeker's Node. It's a simple but important step to ensure smooth operation.\n\nThis request will not trigger a blockchain transaction or cost any gas fees.\n\n🔥 Your node's address: ";
+        string memory prefixLineTwo = unicode"\n\n🆔 Your seeker id: ";
+        string memory prefixLineThree = unicode"\n\n📦 The block this message was signed: ";
+
+        seeker_ownership_message = string(
```
JCSanPedro, line 242: It doesn't make sense to update this field whenever `setSeekerAccount` is called. Wouldn't you have a problem where

1. Account 1 calls setSeekerAccount
2. Account 2 calls setSeekerAccount,
3. Account 1 calls getProofMessage, but the `seeker_ownership_message` field is already set for Account 2?

What you need to do is basically take the implementation for this function `updatePrefix()`, but then it bring it over to `getProofMessage`. The function `getProofMessage` has the parameters `seekerId`, `node`, and `proofBlock`, so you don't need to update any storage values in the contract itself to get the correct message.

JCSanPedro (changes requested):

contracts/Registries.sol
```diff
@@ -227,11 +233,29 @@ contract Registries is Initializable, OwnableUpgradeable {
         return nodes.length;
     }
 
+    function updatePrefix() private {
+        string
+            memory prefixLineOne = unicode"🤖 Hi frend! 🤖\n\n📜 Signing this message proves that you're the owner of this Seeker NFT and allows your Seeker to be used to operate your Seeker's Node. It's a simple but important step to ensure smooth operation.\n\nThis request will not trigger a blockchain transaction or cost any gas fees.\n\n🔥 Your node's address: ";
```
JCSanPedro, line 238: Can you fix the spelling mistake `Hi frend!` -> `Hi friend!`

JCSanPedro: They gave us the message with the spelling mistake but I assume it's suppose to be `friend`

JCSanPedro: Hold on actually, I'll confirm that

test/registries.ts
```diff
@@ -311,4 +327,29 @@ describe('Registries', () => {
...
+      seekerAccount,
+      tokenId,
+    );
+
+    const prefix = await registries.getProofMessage(
+      tokenId,
+      await account.getAddress(),
+      block,
+    );
+    console.log(prefix);
```
JCSanPedro, line 353: You shouldn't just console log it. You should write the test so that it equals what you expect it to be.

test/utils.ts
```diff
@@ -170,11 +170,14 @@ async function setSeekerRegistry(
...
   const accountAddress = await account.getAddress();
-  const proofMessage = `${prefix}:${tokenId}:${accountAddress.toLowerCase()}:${block.toString()}`;
+  const proofMessage = registries.getProofMessage(
+    tokenId,
+    accountAddress,
+    block,
+  );
 
-  const signature = await seekerAccount.signMessage(proofMessage);
+  const signature = await seekerAccount.signMessage(await proofMessage);
```
JCSanPedro, line 180: The `await` call should be done when you call the function. It works either way, but its just convention, and is more readable that way.

contracts/Registries.sol
```diff
@@ -246,16 +236,28 @@ contract Registries is Initializable, OwnableUpgradeable {
         uint256 seekerId,
         address node,
         uint256 proofBlock
-    ) public pure returns (bytes memory) {
+    ) public pure returns (string memory) {
```
JCSanPedro, line 239:
```suggestion
    ) public pure returns (bytes memory) {
```
I don't think you should change this to string otherwise `setSeekerAccount` will have to convert it to bytes. In general, some signing libraries might not work with strings as well, so better to just leave as bytes.

contracts/Registries.sol
```diff
@@ -246,16 +236,28 @@ contract Registries is Initializable, OwnableUpgradeable {
...
-                Strings.toString(proofBlock)
+            string(
+                abi.encodePacked(
+                    PREFIXLINEONE,
+                    Strings.toHexString(uint256(uint160(node)), 20),
+                    PREFIXLINETWO,
+                    Strings.toString(seekerId),
+                    PREFIXLINETHREE,
+                    Strings.toString(proofBlock),
+                    ":",
```
JCSanPedro, line 254: You don't need to keep appending to the string after `Strings.toString(proofBlock)`. You can see that you end up just appending the same values twice.

contracts/Registries.sol
```diff
@@ -246,16 +236,28 @@ contract Registries is Initializable, OwnableUpgradeable {
         uint256 seekerId,
         address node,
         uint256 proofBlock
-    ) public pure returns (bytes memory) {
+    ) public pure returns (string memory) {
+        string
+            memory PREFIXLINEONE = unicode"🤖 Hi frend! 🤖\n\n📜 Signing this message proves that you're the owner of this Seeker NFT and allows your Seeker to be used to operate your Seeker's Node. It's a simple but important step to ensure smooth operation.\n\nThis request will not trigger a blockchain transaction or cost any gas fees.\n\n🔥 Your node's address: ";
```
JCSanPedro, line 241: `PREFIXLINE...` probably doesn't make sense as a name anymore since it's not just a prefix. Tbh I don't think you need to define these constants, it would be more readable to just do it inline.

```
abi.encodePacked(
	unicode"🤖 Hi frend! 🤖\n\n📜 Signing this message proves that you're the owner of this Seeker NFT and allows your Seeker to be used to operate your Seeker's Node. It's a simple but important step to ensure smooth operation.\n\nThis request will not trigger a blockchain transaction or cost any gas fees.\n\n",
	unicode"🔥Your node's address: "
	Strings.toHexString(uint256(uint160(node)), 20),
	unicode"\n\n🆔 Your seeker id: ",
	Strings.toString(seekerId),
	unicode"\n\n📦 The block this message was signed: ",
	Strings.toString(proofBlock),
)
```

JCSanPedro (changes requested):

JCSanPedro (approved):

PR #101, opened 2023-02-15
NET-445 (SLP-01) Fix `revokeSeekerAccount` preventing ticket redemption

JCSanPedro:
## Motivation

The primary restriction regarding Node operators and Seekers is that a Node runner must have registered a valid Seeker with their Node when their Node joins an epoch. There is no restriction that the Node runner must retain ownership of their Seeker after their Node joins an epoch in order for their Node to continue to redeem tickets during that epoch. For example it is valid to sell your Seeker immediately after your Node joins an epoch (there is another check in EpochsManager that would prevent the same seeker being reused).

However there is an additional `require` call in the `redeem` method that checks if the Node's account has a valid Seeker account registered with it. This is unnecessary as the `joinNextEpoch` method already has the validation to ensure the Node's registered Seeker is valid. It also allows a Node runner to accidentally prevent its Node from redeeming tickets if the runner were to call `revokeSeekerAccount`. This is *not* a front running attack as the Seeker account holder and Node runner are assumed to be the same entity.

## Summary of Changes

- Remove the `require` call in `redeem` and remove unit tests for that check

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -3,27 +3,24 @@
 **Protocol and Economic Incentives For Decentralized Communications
 Infrastructure**
 
-Paul Freeman            <paul@sylo.io> </br>
-John Carlo San Pedro    <john@sylo.io> </br>
-Joshua Dawes            <josh@sylo.io> </br>
+Paul Freeman <paul@sylo.io> </br> John Carlo San Pedro <john@sylo.io> </br>
+Joshua Dawes <josh@sylo.io> </br>
 
-## Table of Contents ###########################################
+## Table of Contents
 
-* [Introduction and Background](#introduction-and-background)
-* [Probabilistic Micropayments](#probabilistic-micropayments)
-  * [Micropayment Tickets](#micropayment-tickets)
-* [The Event Relay Protocol](#the-event-relay-protocol)
-  * [Nodes](#nodes)
-* [Staking](#staking)
-  * [Delegated Staking](#delegated-staking)
-  * [Scanning for a node](#scanning-for-a-node)
-* [Epochs](#epochs)
-* [Rewards](#rewards)
-* [Ticket decay with time](#ticket-decay-with-time)
+- [Introduction and Background](#introduction-and-background)
+- [Probabilistic Micropayments](#probabilistic-micropayments)
+  - [Micropayment Tickets](#micropayment-tickets)
+- [The Event Relay Protocol](#the-event-relay-protocol)
+  - [Nodes](#nodes)
+- [Staking](#staking)
+  - [Delegated Staking](#delegated-staking)
+  - [Scanning for a node](#scanning-for-a-node)
+- [Epochs](#epochs)
+- [Rewards](#rewards)
+- [Ticket decay with time](#ticket-decay-with-time)
 
-
-
-## Introduction and Background ###########################################
+## Introduction and Background
 
 Sylo Nodes are an application that anyone can run, to help provide network
 services to Sylo users in a truly private, fully decentralised way.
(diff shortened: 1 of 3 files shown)


JoshD641 (approved):
LGTM. While not a front running attack, this changes fixes a "tokenomics" bug. 

Before this change, a Node whose user had moved their Seeker no longer had an incentive to complete relays that were already in flight, since they would be unable to claim rewards from those relays. Those relays can now be claimed even without a Seeker in the registry, giving the Node an incentive to complete them.

JoshD641 (dismissed):
See previous comment.

docs/overview.md
```diff
@@ -318,6 +318,34 @@ possible, to maximize their income from performing the service. It similarly
 incentivizes them to invest in throughput and uptime improvements, so that they
 can claim tickets as soon as possible, and earn more from the work that they do.
 
+## The Seekers ###########################################
```
teinnt, line 321: The text lgtm, should we try to rewrap this section?

docs/overview.md
```diff
@@ -318,6 +318,34 @@ possible, to maximize their income from performing the service. It similarly
...
+
+It is expected that users will trade/transfer ownership of the Seeker that is being used to run a 
+Node before revoking the link between the Seeker and the Node in the Node's registry entry. This cannot 
+easily be prevented because Seeker trades/transfers are initiated by the Seeker wallet, while registry
+edits are made by the Node's wallet, which does not know of the transfer when it happens.
+
+As a result, the Node is only required to have a Seeker in the registry when it joins an epoch, after which 
+the Node can continue to participate during that epoch even after that Seeker has been transfered to another 
+wallet. This ensures that:
+1) The node can continue to claim rewards on work performed during the epoch, giving it an incentive to continue
```
teinnt, line 342: Should we add a space here: as linter suggestion, lists should be surrounded by blank linesmarkdownlint [MD032](https://github.com/DavidAnson/markdownlint/blob/v0.27.0/doc/md032.md)

docs/overview.md
```diff
@@ -318,6 +318,34 @@ possible, to maximize their income from performing the service. It similarly
...
+As a result, the Node is only required to have a Seeker in the registry when it joins an epoch, after which 
+the Node can continue to participate during that epoch even after that Seeker has been transfered to another 
+wallet. This ensures that:
+1) The node can continue to claim rewards on work performed during the epoch, giving it an incentive to continue 
+performing the service rather than leaving some relays unfulfilled.
+2) The stake directory is static for the duration of the epoch, so that each receiver maps to the same node 
+for the entire epoch.
+However, the Node will be unable to join the following epoch until it registers another Seeker.
+
+
```
teinnt, line 348: Should remove this line?

teinnt (commented):
Just some white space, otherwise LGTM.

docs/overview.md
```diff
@@ -3,27 +3,24 @@
...
+- [Probabilistic Micropayments](#probabilistic-micropayments)
+  - [Micropayment Tickets](#micropayment-tickets)
+- [The Event Relay Protocol](#the-event-relay-protocol)
+  - [Nodes](#nodes)
+- [Staking](#staking)
+  - [Delegated Staking](#delegated-staking)
+  - [Scanning for a node](#scanning-for-a-node)
+- [Epochs](#epochs)
+- [Rewards](#rewards)
+- [Ticket decay with time](#ticket-decay-with-time)
```
JCSanPedro, line 21: Format on save made these changes. Unrelated to the PR but my as well keep them in.

teinnt (approved):
Awesome

PR #102, opened 2023-02-16
Net 447 (SLP-03) Adjust ticket block check

JCSanPedro:
## Motivation

The audit noted that it was possible for a ticket to be redeemed with a generation block greater than the current epoch's end block.

This PR prevents that by additionally checking if the ticket generation block is less than or equal to than the current block.

## Summary of Changes

- Add additional check for ticket generation block
- Remove require check that the ticket generation block is within the current epoch in the `redeem` function. This is because we also perform the same check in `calculateWinningProbability`, so it is unnecessary to do here as well.

### DIFF
--- a/contracts/Payments/Ticketing.sol
+++ b/contracts/Payments/Ticketing.sol
@@ -225,9 +225,8 @@ contract SyloTicketing is Initializable, OwnableUpgradeable {
         EpochsManager.Epoch memory epoch = _epochsManager.getEpoch(ticket.epochId);
         require(epoch.startBlock > 0, "Ticket's associated epoch does not exist");
         require(
-            ticket.generationBlock >= epoch.startBlock &&
-                (epoch.endBlock > 0 ? ticket.generationBlock < epoch.endBlock : true),
-            "This ticket was not generated during it's associated epoch"
+            ticket.generationBlock <= block.number,
+            "The ticket cannot be generated for a future block"
         );
 
         bytes32 ticketHash = getTicketHash(ticket);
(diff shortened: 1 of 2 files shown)


JoshD641 (changes requested):
The code change LGTM. At first I thought it was odd to remove the check from `redeem` rather than from `calculateWinningProbability`, and spread out the checks unnecessarily, but it does makes sense in the context of the ticket win probability decay to do it inside `calculateWinningProbability`.

We should probably leave the old test case in place, rather than editing it, to make sure that the require inside `calculateWinningProbability` is working too.

JCSanPedro:
> We should probably leave the old test case in place, rather than editing it, to make sure that the require inside calculateWinningProbability is working too

There is another test case that already exists for that function specifically https://github.com/dn3010/sylo-ethereum-contracts/blob/master/test/payments/ticketing.ts#L414-L435

JoshD641 (approved):

contracts/Payments/Ticketing.sol
```diff
@@ -225,9 +225,8 @@ contract SyloTicketing is Initializable, OwnableUpgradeable {
         EpochsManager.Epoch memory epoch = _epochsManager.getEpoch(ticket.epochId);
         require(epoch.startBlock > 0, "Ticket's associated epoch does not exist");
```
teinnt, line 226: Do we want to remove the check here, because this will be checked again in `calculateWinningProbability`?

JCSanPedro: Makes sense

teinnt (approved):

PR #103, opened 2023-02-16
NET-455 (SLP-12) Optimize gas

teinnt:
SLP-12 listed out multiple recommendations for optimizing the gas usage of the contracts.

Since each commit represents for one of the requirements, it is suggested to review commit from commit. It will be easier to track and understand the changes.


contracts/Staking/Directory.sol
```diff
@@ -225,11 +227,14 @@ contract Directory is Initializable, OwnableUpgradeable, Manageable {
...
+        address[] memory stakees = new address[](entryLength);
+        uint256[] memory boundaries = new uint256[](entryLength);
 
         DirectoryEntry memory entry;
-        for (uint256 i; i < directories[epochId].entries.length; i++) {
+
+        for (uint256 i; i < entryLength; i++) {
             entry = directories[epochId].entries[i];
             stakees[i] = entry.stakee;
             boundaries[i] = entry.boundary;
```
JCSanPedro, line 247: https://github.com/dn3010/sylo-ethereum-contracts/blob/master/contracts/Payments/Ticketing/RewardsManager.sol#L339-L343 Looks like we can also significantly optimize this loop.

contracts/Epochs/Manager.sol
```diff
@@ -159,26 +165,29 @@ contract EpochsManager is Initializable, OwnableUpgradeable {
         Registries.Registry memory registry = _registries.getRegistry(msg.sender);
 
         // validate the node's seeker ownership
-        require(registry.seekerAccount != address(0), "Seeker account cannot be zero");
+        if (registry.seekerAccount == address(0)) {
+            revert InvalidSeekerAccount("Seeker account cannot be zero", registry.seekerAccount);
```
JCSanPedro, line 169: I don't think we should do this style `error SomeError(string message)`. In this case we can just name the error more specific
`error SeekerAccountNotSet`

Similar for the error when below when checking if the owner of the seeker id matches the seeker account in the registry
`error SeekerOwnerMismatch`

JCSanPedro: Actually I'm pretty sure we won't get any gas cost savings if we end just passing in strings as parameters.

teinnt: Sure 👍

contracts/Staking/Directory.sol
teinnt: I have updated it a bit more here https://github.com/dn3010/sylo-ethereum-contracts/pull/103/commits/7d197111f6ebb5f28c7bfdb4cdaf8df23ad146f9#diff-8f5fbb226f82efbbb08e9b5833fec34612f71254918bf79c9c78c550aafe4ed5R236-R239

JCSanPedro: I see the change is in this PR but I don't see it as being part of that commit. @teinnt Can you amend that commit so it is in that one specifically.

contracts/Payments/Ticketing.sol
```diff
@@ -237,10 +258,9 @@ contract SyloTicketing is Initializable, OwnableUpgradeable {
             ticket.epochId,
             ticket.redeemer
         );
-        require(
-            directoryStake > 0,
-            "Ticket redeemer must have joined the directory for this epoch"
-        );
+        if (directoryStake == 0) {
+            revert RedeemerMustJoinEpoch(ticket.epochId);
```
JoshD641, line 262: Perhaps rename to "rRdeemerMustHaveJoinedEpoch" for clarity?

contracts/Payments/Ticketing/Parameters.sol
```diff
@@ -117,8 +122,26 @@ contract TicketingParameters is Initializable, OwnableUpgradeable {
...
+
+    /**
+     * @notice Retrieve the current ticketing parameters.
+     * @return faceValue The face value of a ticket in SOLO.
+     * @return baseLiveWinProb The base live win probability of a ticket.
+     * @return expiredWinProb The expired win probability of a ticket.
+     * @return decayRate The decay rate of a ticket.
+     * @return ticketDuration The duration of a ticket in number of blocks.
+     */
+    function getTicketingParameters()
```
JoshD641, line 140: There will be some ambiguity in how this is used by the node that we could try to clarify in advance.

Suppose the ticket parameters are updated in Parameters.sol midway through epoch 5 - this function would immediately start returning the new parameters, but epoch 5 will be using the old param values. 

Perhaps this function should also return the first epoch number that these values apply to?

contracts/Registries.sol
```diff
@@ -191,21 +206,23 @@ contract Registries is Initializable, OwnableUpgradeable {
...
-        require(end > start, "End index must be greater than start index");
-        require(
-            end <= nodes.length,
-            "End index cannot be greater than total number of registered nodes"
-        );
+    function getRegistries(
+        uint256 start,
+        uint256 end
+    ) external view returns (address[] memory, Registry[] memory) {
+        uint256 nodeLength = nodes.length;
```
JoshD641, line 213: nodeLength is defined but not used consistently through what follows - doesn't seem important except that the style may as well be consistent

JoshD641 (commented):
A few minor comments to make, but it looks good overall. I'm happy with however you decide to resolve them, so I'll make this review a comment for now and then approve it once you've had a chance to make any changes you think are needed.

contracts/Payments/Ticketing/Parameters.sol
JCSanPedro: I'm not sure if this is needed. The purpose of this function is an optimization so that all parameters can be queried in one call. The problem you pointed out is valid but would still apply if you were to call each getter individually. It is already documented that parameters are constant for the duration of an epoch, and changes are only reflected once a new epoch begins, so we should just rely on that.

JoshD641: Yea, i think it's fine - my main concern was misinterpretation by a 3rd party, but they'd probably be using a library we wrote to get this info rather than calling the contracts directly, so we can always make that interface more "idiot proof" instead of this one.

JoshD641 (approved):

PR #105, opened 2023-02-16
NET-449 (SLP-05) Use SafeCast lib to avoid integer overflow

JCSanPedro:
## Motivation

We perform integer down casts throughout the contract code base. These casts are unsafe as if the value is larger than the maximum value we are casting, then the behaviour of the contract will be unpredictable.

We can use the `SafeCast` which will explicitly throw an error if the value exceeds the maximum. This will prevent unexpected behaviour from happening.

## Summary of Changes

- Replace any integer casts with the `toX` form from the SafeCast lib

### DIFF
--- a/contracts/Payments/Ticketing.sol
+++ b/contracts/Payments/Ticketing.sol
@@ -11,6 +11,7 @@ import "./Ticketing/RewardsManager.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
 import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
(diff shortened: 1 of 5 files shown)


JoshD641 (approved):
LGTM. I used a regex to look for any casts that might have been missed, no issues there.

teinnt (approved):
LGTM. Just wanna confirm, we don't need to cast `uint256` right?

JCSanPedro:
@teinnt We don't need to safe cast for `uint256` as the issue only occurs when downcasting (bigger type going to smaller type). If you check the SafeCast library, there isn't even a `toUint256` function

teinnt:
> * SafeCast

Actually... they have that function... But I understand your point. Thank you John!

teinnt (approved):

JCSanPedro:
> Actually... they have that function... But I understand your point. Thank you John!

They only have that function for `toUint256(int256 value)`, but we are only casting `uint` value in this case.

PR #106, opened 2023-02-16
NET-450 (SLP-06) Improve seeker account signatures

JCSanPedro:
## Motivation

Currently we validate that seeker signatures are submitted within a `proofDuration` time frame. This is to prevent the signature from being reused at a later date. This approach is a remnant of when we were relying on an external Oracle to validate the Seeker proof. This is no longer necessary and we can now just store and use a nonce with each signature to ensure that each one is unique.

## Summary of Changes

- Replace `proofDuration` with a `nonce` in the signature message
- Track nonces in storage to prevent reuse

### DIFF
--- a/contracts/Registries.sol
+++ b/contracts/Registries.sol
@@ -56,6 +56,12 @@ contract Registries is Initializable, OwnableUpgradeable {
      */
     address[] public nodes;
 
+    /**
+     * @notice Tracks nonces used when registering the seeker account
+     * to prevent signature re-use.
+     */
+    mapping(bytes32 => address) private signatureNonces;
+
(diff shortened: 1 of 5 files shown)


JoshD641 (dismissed):
Good approach and good test coverage, LGTM.

teinnt (commented):
Wow, this approach is cool 🚀🚀🚀 Btw, do we wanna update the deploy script and genesis config to too?

JCSanPedro:
> Wow, this approach is cool 🚀🚀🚀 Btw, do we wanna update the deploy script and genesis config to too?

Good catch. Will do now.

teinnt (approved):
LGTM

PR #107, opened 2023-02-16
NET-451 (SLP-07) Improve hashing of sender and redeemer rands

JCSanPedro:
## Motivation

We hash random integers for use as the sender and redeemer commits. The audit identified this as not being secure enough, and suggested we should double hash and use a salt to improve the hash security.

## Summary of Changes

- Modify the commit creation to use the ticket generation block as a salt, and perform `keccak256` twice.

### DIFF
--- a/contracts/Payments/Ticketing.sol
+++ b/contracts/Payments/Ticketing.sol
@@ -312,13 +312,13 @@ contract SyloTicketing is Initializable, OwnableUpgradeable {
         // validate that the sender's random number has been revealed to
         // the redeemer
         require(
-            keccak256(abi.encodePacked(senderRand)) == ticket.senderCommit,
+            createCommit(ticket.generationBlock, senderRand) == ticket.senderCommit,
             "Hash of senderRand doesn't match senderRandHash"
         );
 
         // validate the redeemer has knowledge of the redeemer rand
         require(
-            keccak256(abi.encodePacked(redeemerRand)) == ticket.redeemerCommit,
+            createCommit(ticket.generationBlock, redeemerRand) == ticket.redeemerCommit,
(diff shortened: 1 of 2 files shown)


contracts/Payments/Ticketing.sol
```diff
@@ -312,13 +312,13 @@ contract SyloTicketing is Initializable, OwnableUpgradeable {
         // validate that the sender's random number has been revealed to
         // the redeemer
         require(
-            keccak256(abi.encodePacked(senderRand)) == ticket.senderCommit,
+            createCommit(ticket.generationBlock, senderRand) == ticket.senderCommit,
```
JCSanPedro, line 315: @JoshD641 Using `ticket.generationBlock` which should be larger than `ticket.epochId` and less likely to be reused compared to the epoch Id. Is this secure enough or should we really just add the salt field to the ticket?

JoshD641: Using a salt would definitely be better, but I find it hard to say whether it matters - I don't know enough about how hard it is to reverse a keccak256 hash. 

To my knowledge, it should be nearly impossible to rainbow table a 256 bit hash function, so long as the possible input space is the entire 2^256 combinations of bits - rainbow tables are mostly used against passwords, which have a lot less variability. So I think it mostly depends on how good the random numbers being committed to are -  but I'm not an expert on this stuff.

Does anyone else in the company have a deeper background in this kind of thing? Maybe Felix?

JCSanPedro: I don't know know of anyone specifically, but I'd think the Red4Sec team would have enough knowledge? I'm just relying on their expertise for this one tbh.

JoshD641 (approved):

teinnt (approved):
LGTM

PR #108, opened 2023-02-16
NET 452 (SLP-08) Avoid storage collisions

JCSanPedro:
## Motivation

The protocol contracts rely on the open zeppelin upgradeable contracts to allow us to make changes to the contract after deployment. When writing upgradeable contracts, layout of storage variables must be carefully considered. This is especially true for base contracts.

Currently the `Manageable` contract does not take into consideration that it may also want to add additional fields in the future. This would make it problematic to upgrade later on. The standard approach is to add a `__gap` field https://github.com/dn3010/sylo-ethereum-contracts/blob/master/test/payments/ticketing.ts#L79-L86

## Summary of Changes

- Add a `__gap` field to `Manageable` to allow for additional fields in the future
- Avoid inheriting `OwnableUpgradeable` twice in `RewardsManager` and `Directory` contract. This is a problem noted in a different part of the audit
> Duplicate inheritance
The   and RewardsManager contracts inherit on multiple occasions from , additionally to causing confusion and making the code difficult to read, it is not a recommended practice, since it relies on the compiler and its version used to see how it
interprets this duplicate inheritance.

But is being addressed here as well.


teinnt (commented):
LGTM, do we want to add `__gap` to other contracts that are not `Manageable`?

JCSanPedro:
> LGTM, do we want to add __gap to other contracts that are not Manageable?
It seems like `__gap` is just used for contracts that intend to be inherited from. I reviewed all of our other contracts and I don't believe we will create any contracts that inherit from those contracts anytime in the future.

teinnt (approved):

PR #109, opened 2023-02-16
NET 454 (SLP-11) Verify transfer call

JCSanPedro:
## Motivation

The ERC20 `transfer` and `transferFrom` return `true` when those calls are successful. The contracts should verify those results to prevent unexpected behaviour.

## Summary of Changes

- Use `SafeERC20` for all `transfer` and `transferFrom` calls to verify their execution


teinnt (approved):
Awesome, LGTM!

PR #112, opened 2023-02-17
NET-460 (SLP-17) emit epoch duration

Nick95550:
Editing manager contract to emit event on setting the epoch duration

### DIFF
--- a/contracts/Epochs/Manager.sol
+++ b/contracts/Epochs/Manager.sol
@@ -39,8 +39,6 @@ contract EpochsManager is Initializable, Ownable2StepUpgradeable {
 
     event EpochJoined(uint256 indexed epochId, address indexed node, uint256 indexed seekerId);
 
-    event EpochDurationUpdated(uint256 epochDuration);
-
     Directory public _directory;
 
     Registries public _registries;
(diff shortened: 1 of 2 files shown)


contracts/Epochs/Manager.sol
```diff
@@ -35,6 +35,8 @@ contract EpochsManager is Initializable, OwnableUpgradeable {
 
     event EpochJoined(uint256 epochId, address node, uint256 seekerId);
 
+    event EpochDurationUpdated(uint256 epochDuration);
```
teinnt, line 38: Normally events are declared after variables, should we move them nearby `event NewEpoch`?

teinnt (commented):
This change is looking good, can you also check if any other places needing `emit event`?

test/epochs.ts
```diff
@@ -28,6 +28,10 @@ describe('Epochs', () => {
   });
 
   it('can set epoch duration', async () => {
+    await expect(epochsManager.setEpochDuration(778))
+      .to.emit(epochsManager, 'EpochDurationUpdated')
+      .withArgs(778);
+
     await epochsManager.setEpochDuration(777);
```
JCSanPedro, line 35: Do you need to make this call if you are already making the call above?

JCSanPedro (approved):

PR #115, opened 2023-02-17
NET-461 (SLP-18) Update storage before external call

teinnt:
The Reentrancy attack is a vulnerability that occurs when external contract calls can make new calls to the calling contract, before the initial execution is completed. For a function, this means that the state of the contract could change in the middle of its execution as a result of a call to an untrusted contract or the use of a low-level function with an external address.

It is essential to always make the state changes in the storage before making transfers or calls to external contracts, in addition to implementing the necessary measures to avoid duplicated calls or chain calls to the methods.


contracts/Staking/Manager.sol
```diff
@@ -158,13 +158,13 @@ contract StakingManager is Initializable, OwnableUpgradeable {
...
 
-        // automatically move any pending rewards generated by their existing stake
-        _rewardsManager.updatePendingRewards(stakee, msg.sender);
-
         if (amount == 0) {
             revert CannotStakeZeroAmount();
         }
 
+        // automatically move any pending rewards generated by their existing stake
+        _rewardsManager.updatePendingRewards(stakee, msg.sender);
```
JCSanPedro, line 166: Seems like we move this external call to the end of the funciton?

teinnt: Yeah, I got those errors when moved it end of the function, so I just moved it after the check.
Screen Shot 2023-02-21 at 7 09 59 AM

JCSanPedro: Ah right, of course. The rewards calculation depends on the current stake.

JCSanPedro (approved):

PR #118, opened 2023-02-21
NET-463 (SLP-20) Validate input

teinnt:
Inputs to functions should be validated to prevent any unexpected behaviour. 

Some examples:

- Methods that take in addresses for contracts should be validated that they not address(0). Can also use `supportsInterface`.
- Range validation for certain values (proofDuration for Registries contract)
- Compute ticket hash may be unnecessary as the hash is never validated

### DIFF
--- a/contracts/Registries.sol
+++ b/contracts/Registries.sol
@@ -19,8 +19,6 @@ import "./interfaces/IRegistries.sol";
 contract Registries is IRegistries, Initializable, Ownable2StepUpgradeable, IERC165 {
     using ECDSA for bytes32;
 
-    uint256 private constant MAX_SEEKER_ID = 47894;
-
     /**
      * @notice ERC721 contract for bridged Seekers. Used for verifying ownership
      * of a seeker.
@@ -65,7 +63,6 @@ contract Registries is IRegistries, Initializable, Ownable2StepUpgradeable, IERC
 
     event DefaultPayoutPercentageUpdated(uint16 defaultPayoutPercentage);
 
-    error SeekerIdOutOfRange();
     error NonceCannotBeReused();
     error EndMustBeGreaterThanStart();
     error PercentageCannotExceed10000();
@@ -76,8 +73,6 @@ contract Registries is IRegistries, Initializable, Ownable2StepUpgradeable, IERC
     error RootSeekersCannotBeZeroAddress();
     error SeekerAccountCannotBeZeroAddress();
     error EndCannotExceedNumberOfNodes(uint256 nodeLength);
-
-    error Test(bytes4 id);
 
     function initialize(
         IERC721 rootSeekers,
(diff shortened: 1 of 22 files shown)


contracts/Registries.sol
```diff
@@ -80,29 +65,45 @@ contract Registries is Initializable, Ownable2StepUpgradeable {
...
     error PercentageCannotExceed10000();
     error PublicEndpointCannotBeEmpty();
     error SeekerAccountMustOwnSeekerId();
     error SeekerAccountMustBeMsgSender();
     error ProofNotSignedBySeekerAccount();
+    error RootSeekersCannotBeZeroAddress();
+    error SeekerAccountCannotBeZeroAddress();
     error EndCannotExceedNumberOfNodes(uint256 nodeLength);
 
+    error Test(bytes4 id);
```
JCSanPedro, line 80: Did you intend to leave this here?

contracts/Registries.sol
```diff
@@ -4,37 +4,22 @@ pragma solidity ^0.8.18;
...
-        //  { name: string, multiaddrs: string[] }
-        string publicEndpoint;
-        // The account which owns a seeker that will be used to
-        // operate the Node for this registry.
-        address seekerAccount;
-        // The id of the seeker used to operate the node. The owner
-        // of this id should be the seeker account.
-        uint256 seekerId;
-    }
+    uint256 private constant MAX_SEEKER_ID = 47894;
```
JCSanPedro, line 22: Tbh I don't think we need to validate this. The fact that up 47895 seekers were minted is just coincidental to the protocol, and doesn't actually impact the behaviour. It also makes it easier to test using the mock TestSeekers contract as we can essentially register any seeker id value we want.

contracts/payments/SyloTicketing.sol
```diff
@@ -105,7 +98,46 @@ contract SyloTicketing is Initializable, Ownable2StepUpgradeable {
         RewardsManager rewardsManager,
         uint256 _unlockDuration
     ) external initializer {
+        if (address(token) == address(0)) {
```
JCSanPedro, line 101: I think we should also do a supportsInterface check for the ERC20 token address as well. But let's double check that this works on Porcini as well for the bridged token.

JCSanPedro:
I think this PR also didn't address the `Compute ticketHash` problem. It suggests that we should call `getTicketHash` in `requireValidWinningTicket` to avoid the need to pass it in as a param. This will make that function easier to use.

teinnt:
> requireValidWinningTicket

Hmm, I think I did it already [here](https://github.com/dn3010/sylo-ethereum-contracts/pull/118/files#diff-dc56f34d3e1de3cca8170dea444190340348add089e2b57385342f9817c22f12R394)? Or did I miss something?

JCSanPedro:
> > requireValidWinningTicket
> 
> Hmm, I think I did it already [here](https://github.com/dn3010/sylo-ethereum-contracts/pull/118/files#diff-dc56f34d3e1de3cca8170dea444190340348add089e2b57385342f9817c22f12R394)? Or did I miss something?

Oh yes, sorry missed it :)

JCSanPedro (approved):

PR #119, opened 2023-02-21
Net 456 (SLP-13) Safe ownership transfer

JCSanPedro:
## Motivation

We use the `Ownable` lib to allow the Sylo to administrate the network protocol. This is an industry standard but imposes some risk in the case we accidentally transfer to an account that is invalid or does not exist. This PR replaces the Ownable contract with `Ownable2Step`, which forces the transferee of the ownership to explicitly send a claim for the ownership in order for the ownership to transer.

The `Ownable2Step` interface https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/access/Ownable2StepUpgradeable.sol

## Summary of Changes

- Replace `Ownable` with `Ownable2Step`

### DIFF
--- a/contracts/ECDSA.sol
+++ b/contracts/ECDSA.sol
@@ -84,3 +84,4 @@ library ECDSA {
         return signer;
     }
 }
+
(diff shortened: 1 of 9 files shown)


teinnt (approved):
LGTM

PR #120, opened 2023-02-21
NET-444 Optimise event ticket validation in notify events

Helenpham1097:
## Motivation

In phase 3, the node will check the senders' deposits when they request relays to the node. Instead of making a call to the blockchain for each relay request that is highly expensive in terms of API call rate, the nodes will store senders' deposits in their own caches, and just need to query on chain after an unlock duration to update data.

However, we have to make sure that the information of senders' deposit in cache must be up-to-date, so that the node can eliminate from being exploited by some bad senders. 

There are an issue that might negatively affect the node

**Problem ** 
The senders' deposits might be burned by other nodes after this node saves the valid senders' deposits in its cache. Because this node depends on its cache, (just query again after the unlock_duration time), this node may not know that these senders' deposit are invalid on chain right now. Therefore, when this node accept relays from these senders, it cannot redeem the winning ticket. 
**Solution**
We can add SenderPenaltyBurnt event when senders' penalty are burned. All nodes can subscribe to this event. When this happens, all the node can know and filter the bad senders in their own cache.

## Summary of changes

- add event when sender's penalty is burned

### DIFF
--- a/contracts/interfaces/payments/ISyloTicketing.sol
+++ b/contracts/interfaces/payments/ISyloTicketing.sol
@@ -5,7 +5,6 @@ interface ISyloTicketing {
     struct Deposit {
         uint256 escrow; // Balance of users escrow
         uint256 penalty; // Balance of users penalty
-        uint256 callUnlockAt; // Block number a user call unlock process
         uint256 unlockAt; // Block number a user can withdraw their balances
     }
 
--- a/contracts/interfaces/payments/ISyloTicketing.sol
+++ b/contracts/interfaces/payments/ISyloTicketing.sol
@@ -5,6 +5,7 @@ interface ISyloTicketing {
     struct Deposit {
         uint256 escrow; // Balance of users escrow
         uint256 penalty; // Balance of users penalty
+        uint256 callUnlockAt; // Block number a user call unlock process
         uint256 unlockAt; // Block number a user can withdraw their balances
     }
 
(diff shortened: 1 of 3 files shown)


contracts/Payments/Ticketing.sol
```diff
@@ -14,22 +14,22 @@ import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
...
+    error UnlockingNotCompleted();
+    error TicketAlreadyRedeemed();
+    error InvalidTicketSignature();
+    error TicketNotCreatedInTheEpoch();
+    error TicketSenderCannotBeZeroAddress();
+    error TicketRedeemerCannotBeZeroAddress();
+    error TicketCannotBeFromFutureBlock();
+    error RedeemerMustHaveJoinedEpoch(uint256 epochId);
+    error SenderCommitMismatch();
+    error RedeemerCommitMismatch();
```
JCSanPedro, line 32: Should remove these whitespace additions from the PR

contracts/Payments/Ticketing.sol
```diff
@@ -40,6 +40,7 @@ contract SyloTicketing is Initializable, Ownable2StepUpgradeable {
     struct Deposit {
         uint256 escrow; // Balance of users escrow
         uint256 penalty; // Balance of users penalty
+        uint256 callUnlockAt; //Block number a user call unlock
```
JCSanPedro, line 43: What is this field used for?

JCSanPedro (changes requested):

contracts/Payments/Ticketing.sol
Helenpham1097: @JCSanPedro Hi John, I just add explanation in the PR message. Please have a look at this. Thank you

JCSanPedro (approved):

PR #121, opened 2023-02-22
General Maintenance

JCSanPedro:
Some misc changes.

* Remove outdated diagram from docs
* Switch to `yarn` for tracking deps
* Run `prettier` command


teinnt (commented):
The change LGTM, should we add a command for running `prettier` in `package.json` and run through the whole project?

JCSanPedro:
> The change LGTM, should we add a command for running `prettier` in `package.json` and run through the whole project?

I think we can add a git commit hook to run the prettier command each time. For now, it's just a nice to have.

PR #122, opened 2023-02-27
NET-466 Update audit to node

teinnt:
The recent protocol contract audit introduced a number of changes that also needs to be reflected in the network client and sylo-node repos.


scripts/package_abis.ts
```diff
@@ -15,6 +15,9 @@ const syloContracts = [
   'SyloTicketing',
   'SyloToken',
   'TicketingParameters',
+  'SyloUtils',
+  'Manageable',
+  'ECDSA',
```
JCSanPedro, line 20: Do we need to package this abi?

contracts/libraries/Manageable.sol
```diff
@@ -3,6 +3,9 @@ pragma solidity ^0.8.18;
 
 import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
 
+error OnlyManagers();
+error ManagerCannotBeZeroAddress();
```
JCSanPedro, line 7: Why do we need to move this from outside the interface?

JCSanPedro (approved):

PR #126, opened 2023-03-06
NET-465 update docker image

Nick95550:
The ethereum contracts repo holds a Dockerfile that builds an image which runs a local ganache testing network. This Dockerfile deploys the contracts to the ganache testnet so they are available when running the image. 

This is used when running the testnet from the so sylo node repo. To make testing more convenient the protocol should be setup in a state where a node can be used to start redeeming tickets. This PR initialises the testnet state so the deployer address can begin generating rewards by redeeming tickets.


scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+  },
+  {
+    privateKey: NodeData.nodeThreePK,
+    publicEndpoint: NodeData.nodeThreePublicEndPoint,
+  },
+  {
+    privateKey: NodeData.nodeFivePK,
+    publicEndpoint: NodeData.nodeFivePublicEndPoint,
+  },
+];
```
JCSanPedro, line 45: I think a block of code like this can be significantly improved by using a loop, or just structure the NodeData json better. I would suggest this
```
nodes.json

{
  "deployerPK": "0x150934096e7bcd0485d154edd771b4466680038a068ccca8e8b483dce8527245",
  "incentiveNodePK": "0x22194c49f4362a63de672f60ea4b5f5166f55a4a231c10c2895f32241b7d4585",
  "relayNodes": [
    {
      "privateKey": "0x98bf857054c6aa0e92ddb6b98a3b63c603261be18b9ca7e882e7a95f297d3edb",
      "publicEndpoint": "http://localhost/28901/public/metadata"
    },
    {
      "privateKey": "0x17bb9893a53d58f4ea4f5d5f1248711d5e722be96c1f81ac562f815af078ecd0",
      "publicEndpoint": "http://localhost/28903/public/metadata"
    },
    {
      "privateKey": "0x2e445c0c320c158d9e3b40e8a419c7960dfe44225dc248eeb5b3bb7ac1a575e8",
      "publicEndpoint": "http://localhost/28905/public/metadata"
    },
    {
      "privateKey": "0x01a1b704b2efb0be843e9918b523da88baea76fcaf6e2b23ac1e3220b6b3523c",
      "publicEndpoint": "http://localhost/28907/public/metadata"
    }
  ]
}
```

See how the `relayNodes` field already matches the type `NodeConfig[]`. When you import that json, you can use it immediately.
```
import Nodes from 'node.json';
...

for (const nodeConfig of Nodes.relayNodes) {
    nodeList[i] = await createNode(provider, nodeConfig);
}
```

scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+
+  const provider = new ethers.providers.JsonRpcProvider('http://0.0.0.0:8545');
+
+  const contracts = await conectContracts(provider);
+
+  for (let i = 0; i < nodeConfigs.length; i++) {
+    nodeList[i] = await createNode(provider, nodeConfigs[i]);
+  }
+
+  const deployerAccount = nodeList[1];
```
JCSanPedro, line 68: Since we are running this script with hardhat, the deployer account would usually be acquired like this
```
import { ethers } from 'hardhat';
...
const [deployer] = await ethers.getSigners();
```

Example https://github.com/dn3010/sylo-ethereum-contracts/blob/master/deploy/00_deploy_phase_two.ts#L29

Hardhat will plugin a list of signers based on this value https://github.com/dn3010/sylo-ethereum-contracts/blob/master/hardhat.config.ts#L33. The deployer account is the first signer.

NOTE: You have to import `ethers` from `hardhat` as opposed to the `ethers` package directly.

scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+    token,
+    stakingManager,
+    seekers,
+    registries,
+    contractAddress,
+    ticketing,
+    ticketingParameters,
+    epochsManager,
+  } as Contracts;
+}
```
JCSanPedro, line 147: This function could probably moved to its own file `scripts/utils.ts`

scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+    publicEndPoint: nodeConfig.publicEndpoint,
+  };
+}
+
+async function addStake(
+  contracts: Contracts,
+  nodes: ethers.Signer,
+  deployer: ethers.Signer,
+): Promise<void> {
+  if (nodes.getAddress() == deployer.getAddress()) {
```
JCSanPedro, line 169: The relay node account should be different from the deployer account, so we shouldn't need this check

scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+    await contracts.token
+      .connect(deployer)
+      .approve(contractAddress.stakingManager, 90000000000000);
+    await contracts.stakingManager
+      .connect(nodes)
+      .addStake(100000, nodes.getAddress());
+  } else {
+    await contracts.token
+      .connect(deployer)
+      .transfer(nodes.getAddress(), 1100000);
```
JCSanPedro, line 179: Better to do the sylo token transfer outside of this function I think

scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+    await contracts.token
+      .connect(nodes)
+      .approve(contractAddress.stakingManager, 90000000000000);
+    await contracts.stakingManager
+      .connect(nodes)
+      .addStake(100000, nodes.getAddress());
```
JCSanPedro, line 185: So this relates to the topic regarding decimal places. Here you are only adding `0.0000000000001` SYLO. If you want to add 100K SYLO, you need to specify `100000000000000000000000`. This is obviously cumbersome, so I recommend using the `ethers.utils.parseEther` fn https://docs.ethers.org/v4/api-utils.html.

```
await contracts.stakingManager
      .connect(nodes)
      .addStake(ethers.utils.parseEther('100000', nodes.getAddress());
```

This function will automatically multiply the input value by `10**18`

JCSanPedro: I think you need to update this everywhere in the script.

scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+    .approve(contractAddress.ticketing, 1100000);
+  await contracts.ticketing
+    .connect(incentivisedNode)
+    .depositEscrow(1000000, incentivisedNode.getAddress());
+  await contracts.ticketing
+    .connect(incentivisedNode)
+    .depositPenalty(100000, incentivisedNode.getAddress());
+}
+
+async function initEpoch(contracts: Contracts, deployer: ethers.Signer) {
```
JCSanPedro, line 251:
```suggestion
async function initNetwork(contracts: Contracts, deployer: ethers.Signer) {
```

JCSanPedro: Seems like this function is doing more than just initializing the epoch, so a different name is more appropriate.

scripts/init_network.ts
```diff
@@ -0,0 +1,279 @@
...
+
+export function connectSigner(
+  wallet: ethers.Wallet,
+  provider: ethers.providers.Provider,
+): ethers.Wallet {
+  const s = wallet.connect(provider);
+
+  const sendTx = s.sendTransaction.bind(s);
+
+  s.sendTransaction = async t => {
```
JCSanPedro, line 269: No need to add this stuff to this script. This was just a temporary work around we did for testing on porcini as it had a bug with gas limits.

You can just call `wallet.connect(provider)` and that's it.

JCSanPedro (changes requested):

scripts/init_network.ts
Nick95550: Hm i'm not sure what you mean by this. This makes sure the deployer account doesn't transfer funds to itself.

JCSanPedro (changes requested):

JCSanPedro (approved):

PR #128, opened 2023-03-13
NET-484 Fix used tickets reentrancy problem

JCSanPedro:
The second audit review noted that the movement of `usedTickets[ticketHash] = true` presented a vulnerability as it ended up being placed after an external call.

This PR reverts that change.


teinnt (approved):
LGTM

PR #130, opened 2023-04-04
NET-508 Improve testing with sylo node testnet

Helenpham1097:
### Motivation

- To improve testing with the go-sylo-node testnet, the init_network script has established multiple network parameters, as well as setting up the stakes and seekers for multiple nodes. However, it hasn't had node join the first epoch and first epoch has not initialized (actually, the order nodes' join and init epoch are wrong. 

Changing the order will avoid the need to manually to connect to a node to join the next epoch, then connect to an incentivising node to start epoch.

- The block will be increased only if the testnet receives pending transaction only, otherwise, it will not increase.
To mimic the blockchain, we add a script to send a post request to mined a new block every 10 seconds. It will be necessary for testing.

### Summary of changes

- change order epoch join and init in the init_network.sh to make it more convenient for testing
- add init network to increase blocks in testnet every 10 seconds


docker/ganache-with-protocol/Dockerfile
```diff
@@ -28,12 +28,19 @@ COPY docker/ganache-with-protocol/deploy_contracts.sh .
...
+
 FROM trufflesuite/ganache:v7.7.6
 
 WORKDIR /app
 COPY --from=node-builder /app/ganache-data /app/ganache-data
 COPY --from=node-builder /app/deployment /app/deployment
+COPY --from=node-builder /app/start_network.sh /app/start_network.sh
+
+#install curl package
+RUN apt-get update && apt-get install -y curl
```
JCSanPedro, line 42: Installation of tools/libs usually happens near the top of the Dockerfile

docker/ganache-with-protocol/start_network.sh
```diff
@@ -0,0 +1,15 @@
...
+    curl -s -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"evm_mine","params":[],"id":1}' http://localhost:8545
+    sleep 10
+  done
+}
+
+# mine block in background
+run_evm_mine &
+
+## run ganache node
+node /app/dist/node/cli.js "$@"
```
JCSanPedro, line 15: Missing new line at the end?

scripts/init_network.ts
```diff
@@ -61,6 +44,24 @@ async function main() {
...
+    await contracts.token
+        .connect(deployer)
+        .transfer(
+            node.signer.getAddress(),
+            ethers.utils.parseEther('1000000000'),
+        );
+
+    await registerNodes(contracts, node);
+    await depositTicketing(contracts, node.signer);
+    await contracts.epochsManager.connect(node.signer).initializeEpoch();
```
JCSanPedro, line 61: `initializEpoch` shouldn't be called in this loop, otherwise for every node, you would try to init the next epoch. This call should happen just after the loop.

JCSanPedro (approved):

JCSanPedro (dismissed):

teinnt:
Looks like we can use `--miner.blockTime 10` as a command in the docker compose file to auto increase the block time. I think we just need to use it instead of calling `evm_mine`.

teinnt (approved):
LGTM

PR #134, opened 2023-05-10
NET-541 Update public endpoint https

teinnt:
Update `nodes.json` info to to match the changes from go-sylo-node testnet https://github.com/dn3010/go-sylo-node/pull/383


teinnt:
This PR is not needed anymore as the [TLS feature](https://github.com/dn3010/go-sylo-node/pull/383) is closed

PR #135, opened 2023-05-17
NET-552 Add receiver signature when redeem ticket

teinnt:
Currently the ticket redemption process is protected by having the receiver decrypt a portion of the ticket, then returning that portion to the relay node. This problems with this process is outlined here [https://www.notion.so/futureverse/e34604acc8ad423e88da2d3de438a668?v=6acf6334375c4c698c7f97bbb8f84508&p=c4 bed91907b348d7a4c46cd885b5665a&pm=s](https://www.notion.so/futureverse/e34604acc8ad423e88da2d3de438a668?v=6acf6334375c4c698c7f97bbb8f84508&p=c4bed91907b348d7a4c46cd885b5665a&pm=s)

This task involves adding an additional field to the ticket structure to indicate the receiver’s signature.


contracts/interfaces/payments/ISyloTicketing.sol
```diff
@@ -11,7 +11,9 @@ interface ISyloTicketing {
     struct Ticket {
         uint256 epochId; // The epoch this ticket is associated with
         address sender; // Address of the ticket sender
-        address delegatedAccount; // Address of the ticket's signer if not the original sender (optional)
+        address senderDelegatedAccount; // Address of the ticket's signer if not the original sender (optional)
+        address receiver; // Address of the ticket receiver
+        address receiverDelegatedAccount; // Address of the ticket's signer if not the original receiver (optional)
```
JCSanPedro, line 16: I think we can create a specific type that wraps a user and their delegated account.

```
struct User {
  address main;
  address delegated;
}
```

Can define in `ISyloTicekting.sol` for now

contracts/payments/SyloTicketing.sol
```diff
@@ -352,14 +363,32 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
...
+        IAuthorizedAccount.Permission permission = IAuthorizedAccount.Permission.DepositWithdrawal;
+        // check sender delegated account
+        if (ticket.senderDelegatedAccount != address(0)) {
+            if (
+                !_authorizedAccount.validatePermission(
+                    ticket.sender,
+                    ticket.senderDelegatedAccount,
+                    permission
+                )
+            ) {
```
JCSanPedro, line 375: I think we can wrap this logic in a function
```
function hasValidDepositPermission(User user) internal (bool) {
  ...
}
```

contracts/payments/SyloTicketing.sol
```diff
@@ -433,11 +466,24 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
...
         }
 
-        if (!isWinningTicket(sig, ticket, senderRand, redeemerRand)) {
+        if (
+            !isValidTicketSig(
+                receiverSig,
+                ticket.receiver,
+                ticket.receiverDelegatedAccount,
+                ticketHash
+            )
```
JCSanPedro, line 481: DId you run prettier? Maybe just also seperate the args in the call `isValidTicketSig(senderSig, ticket.sender, ticket.senderDelegatedAccount, ticketHash)` so we have format consistency? Unless prettier is doing this automatically

JCSanPedro (changes requested):
Good work 👍

contracts/payments/SyloTicketing.sol
teinnt: Yeah, this one is auto reformatter by Prettier, because the length is longer than the default length.

contracts/payments/SyloTicketing.sol
```diff
@@ -67,8 +67,9 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
     event SenderPenaltyBurnt(address sender);
     event Redemption(
         uint256 indexed epochId,
-        address indexed sender,
         address indexed redeemer,
+        address indexed sender,
+        address receiver,
```
teinnt, line 72: The maximum event index is 3 so we cannot have `receiver` as an indexed parameters.

contracts/payments/SyloTicketing.sol
```diff
@@ -422,26 +423,49 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
...
             revert RedeemerCommitMismatch();
         }
 
-        if (!isValidTicketSig(sig, ticket.sender, ticket.delegatedAccount, ticketHash)) {
-            revert InvalidTicketSignature();
+        if (!hasValidDepositPermission(ticket.sender)) {
+            revert SenderDelegatedAccountDoesNotHaveWithdrawalPermission();
+        }
+        if (!hasValidDepositPermission(ticket.receiver)) {
+            revert ReceiverDelegatedAccountDoesNotHaveWithdrawalPermission();
```
JCSanPedro, line 435: I'm thinking `DepositWithdrawal` isn't the best name now that we have the receiver signature as well.

Maybe change to `TicketSigning`?

contracts/payments/SyloTicketing.sol
```diff
@@ -422,26 +423,49 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
...
+        ) {
+            revert InvalidSenderSignature();
+        }
+
+        if (
+            !isValidTicketSig(
+                receiverSig,
+                ticket.receiver.main,
+                ticket.receiver.delegated,
+                ticketHash
```
JCSanPedro, line 449: I think we can replace these two parameters with just one `User` parameter

test/payments/ticketing.test.ts
```diff
@@ -687,10 +684,22 @@ describe('Ticketing', () => {
       .connect(aliceConnected)
       .authorizeAccount(delegatedWallet.address, permission);
 
-    const { ticket, senderRand, redeemerRand, signature } =
-      await createWinningTicket(alice, owner, undefined, delegatedWallet);
+    const { ticket, redeemerRand, senderSig, receiverSig } =
+      await createWinningTicket(
+        alice,
+        alice,
```
JCSanPedro, line 690: I think for the tests we should create a different account for the receiver

JCSanPedro (dismissed):
Good work 👍 .

@JoshD641 can you review please.

JoshD641 (dismissed):
The code changes look good to me =)

This might be related to a different PR - It'd be good to record the conversations we've had about blacklisting, and whether the node can force tickets to win by setting the rand, on notion. Moving away from using the sender's rand, to using the receiver signature, in the ticket logic is a fairly big change, and writing down the implications of that change will be good context for later. 

It might makes sense to update and extend this doc, which currently refers to a version of the code that uses both rands: https://www.notion.so/futureverse/RFC-Avoiding-reliance-on-asymmetric-encryption-for-protecting-ticket-redemptions-c4bed91907b348d7a4c46cd885b5665a

(I can have a go at it if you think it's a good idea @JCSanPedro )

contracts/AuthorizedAccount.sol
```diff
@@ -190,16 +224,34 @@ contract AuthorizedAccount is IAuthorizedAccount, Initializable, Ownable2StepUpg
...
+
+        for (uint i = 0; i < authAccounts.length; ++i) {
+            if (authAccounts[i].account == authorized) {
+                for (uint j = 0; j < authAccounts[i].permissions.length; ++j) {
+                    if (authAccounts[i].permissions[j].permission == permission) {
+                        uint256 authorizedAt = authAccounts[i].permissions[j].authorizedAt;
+                        uint256 unauthorizedAt = authAccounts[i].permissions[j].unauthorizedAt;
+
+                        // authorizedAt <= atBlock < unauthorizedAt
+                        if (authorizedAt <= atBlock && atBlock < unauthorizedAt) {
```
JoshD641, line 241: I think this section could be expressed in a way that's easier to read. 

It seems to be correct, but the logic relies on the default value of unauthorisedAt, which made it harder to understand what idea the checks were capturing conceptually without jumping around the code.

Maybe something like this?

if (wasUnAuthorisingAtBlock || wasAuthorisedAtBlock ){
return true
}

Where those bools are calculated separately on earlier lines (using the same logic you've used in the two if statements), would make it a bit faster to read.

contracts/interfaces/IAuthorizedAccount.sol
```diff
@@ -3,13 +3,19 @@ pragma solidity ^0.8.18;
...
 interface IAuthorizedAccount {
     enum Permission {
-        DepositWithdrawal
+        TicketSigning
+    }
+
+    struct AuthorizedPermission {
+        Permission permission; // Permission
+        uint256 authorizedAt; // Block number the permission is last authorized (block.number)
+        uint256 unauthorizedAt; // Block number the permission is started to be unauthorized (block.number + 1)
```
JoshD641, line 12: "Block number at which the permission was last set to authorised" is probably more correct for the first comment. 

For the unauthorised comment, does this mean that the permission is considered to be live during block unauthorisedAt, and not live during block unauthorisedAt +1 ?

contracts/interfaces/IAuthorizedAccount.sol
```diff
@@ -3,13 +3,19 @@ pragma solidity ^0.8.18;
...
     }
 
     struct AuthorizedAccount {
-        address account; // Address of the authorized account
-        uint256 createdAt; // Block number the authorized account was created
-        Permission[] permissions; // Permissions associated with delegated account
+        address account; // Authorized account
+        uint256 authorizedAt; // Block number the main account authorized the authorized account, 0 if not authorized
```
JoshD641, line 17: I can see potential for misunderstand with the phrasing of "0 if not authorised". 

Are you saying that it's 0 if it has never been authorised in the past, or 0 if it is currently not authorised?

Based on the code i think the former is true, but it reads like the latter to me.

test/authorizedAccount.test.ts
```diff
@@ -36,366 +38,783 @@ describe('Authorized Account', () => {
...
+      delegatedAccount1,
+      Permission.TicketSigning,
+      1000,
+    );
+    assert.equal(validate, true);
+
+    // A < B < U => true
+    await utils.advanceBlock(5);
+    await authContract.unauthorizeAccount(delegatedAccount1);
+    let unauthBlock = (await currentBlock()) + 1; // unauthorizeAt = block.number + 1
```
JoshD641, line 691: Is there any way that these two lines could be run during different blocks, leading to unauthBlock being not the same as the unauthBlock in the contract's memory? I'm not sure how the async works with await in this case.

JoshD641 (commented):
The PR looks good as far as I can see - I identified issues around comments and readability, and the tests cover every case I could think of. The only reason this isn't an "approve" review is related to my question in the comment on the tests - in case that creates a situation where the test isn't testing what we think it is.

test/authorizedAccount.test.ts
teinnt: I think it is fine because Hardhat test does not auto mint new block by default, a new block is created when a new transaction is called.

contracts/AuthorizedAccount.sol
```diff
@@ -190,19 +231,58 @@ contract AuthorizedAccount is IAuthorizedAccount, Initializable, Ownable2StepUpg
...
+
+                        if (isPermissionUnauthorized(authorizedAt, unauthorizedAt)) {
+                            return
+                                isAtBlockBetweenAuthorizedDuration(
+                                    atBlock,
+                                    authorizedAt,
+                                    unauthorizedAt
+                                );
+                        }
+                        return isAtBlockGreaterThanAuthorizedAt(atBlock, authorizedAt);
```
JCSanPedro, line 255: I don't think it is good practice to create functions that execute just a condition check, especially since these are only being used once. It means as a reader we would have to verify the function implementation as well. I think it would be more sensisble to use a combination of well named variables and comments to make the code more readable.

```
bool isPermissionUnauthorized = authorizedAt > 0 && authorizedAt < unauthorizedAt;
if (isPermissionUnauthorized) {
  // the permission was previously valid, so we check that
  // the `atBlock` is referencing a time when the permission was valid
  return authorizedAt <= atBlock && atBlock < unauthorizedAt;
}

// Otherwise just check if the permission was authorized before the
// atBlock
return authorizedAt > 0 && authorizedAt <= atBlock;
```

contracts/interfaces/IAuthorizedAccount.sol
```diff
@@ -3,13 +3,51 @@ pragma solidity ^0.8.18;
...
+     * is updated, because they are both needed when validating the permission.
+     */
+    struct AuthorizedPermission {
+        // Permission type
+        Permission permission;
+        // Block number from which the permission was set to be authorised.
+        // If the transaction is called in block 1, the permission is
+        // authorized from block 1 (authorizedAt = block.number).
+        uint256 authorizedAt;
+        // Block number from which the permission was set to be unauthorised.
```
JCSanPedro, line 26:
```suggestion
        // Block number from which the permission was set to be unauthorized.
```

We should try be consistent for the spelling convnetion

JCSanPedro (changes requested):

JCSanPedro (approved):

PR #136, opened 2023-05-17
NET-550 Add-delegated-account

Helenpham1097:
## Motivation
As outlined in this here https://www.notion.so/futureverse/RFC-Alleviating-user-friction-in-ticket-creation-process-1b1416776aee4532b513c638ae976e23#1616abd90910416da6fc434fe41e5e0a, we can modify the signature verification check for tickets to rely on a signature from a different wallet, one that has been delegated by the holder of the deposit.

This solution involves:
- Implement new contract to store authorised accounts
- Add additional field in current ticket structure to indicate if the signature is from the main account or a delegated account
- Check permission of the delegated account before withdrawing sender deposit
- Verify signature if delegated account is used in the ticket

## Summary of changes 
- implement authorized account contract
- integrate authorized account contract with ticketing contract
- adding necessary tests

### DIFF
--- a/contracts/AuthorizedAccount.sol
+++ b/contracts/AuthorizedAccount.sol
@@ -1,0 +1,220 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+
+import "./interfaces/IAuthorizedAccount.sol";
+
+/**
+ * @notice Manages authorized accounts with limited permissions on behalf of main account
+ * these authorized accounts are allowed to perform some certain actions in the Sylo network
+ * in order to reduce the works for main account
+ */
+contract AuthorizedAccount is IAuthorizedAccount, Initializable, Ownable2StepUpgradeable, ERC165 {
+    /**
+     * @notice Tracks authorized accounts for every main account
+     */
+    mapping(address => AuthorizedAccount[]) public authorizedAccounts;
+
+    error AuthorizedAccountCannotBeZeroAddress();
+    error MainAccountCannotBeZeroAddress();
+    error AccountAlreadyExists();
+    error AccountDoesNotExist();
+
+    function initialize() external initializer {
+        Ownable2StepUpgradeable.__Ownable2Step_init();
+    }
+
+    /**
+     * @notice Returns true if the contract implements the interface defined by
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(IAuthorizedAccount).interfaceId;
+    }
+
+    /**
+     * @notice Adds new authorized accounts with certain permissions. This will revert if the account
+     * has already existed
+     * @param authorized The authorized address of the main account
+     * @param permissions The list of permissions that the authorized account
+     * can perform within the Sylo network.
+     */
+    function authorizeAccount(address authorized, Permission[] calldata permissions) external {
+        if (authorized == address(0)) {
+            revert AuthorizedAccountCannotBeZeroAddress();
+        }
+
+        // check if account has already existed
+        AuthorizedAccount[] storage _authorizedAccounts = authorizedAccounts[msg.sender];
+        for (uint i = 0; i < _authorizedAccounts.length; i++) {
+            if (_authorizedAccounts[i].account == authorized) {
+                revert AccountAlreadyExists();
+            }
+        }
+
+        AuthorizedAccount memory newAccount = AuthorizedAccount({
+            account: authorized,
+            createdAt: block.number,
+            permissions: permissions
+        });
+
+        authorizedAccounts[msg.sender].push(newAccount);
+    }
+
+    /**
+     * @notice Removes an authorized account associated with the msg.sender.
+     * This will revert if the account does not exist
+     * @param authorized The address of the authorized account
+     */
+    function unauthorizeAccount(address authorized) external {
+        if (authorized == address(0)) {
+            revert AuthorizedAccountCannotBeZeroAddress();
+        }
+
+        AuthorizedAccount[] storage _authorizedAccounts = authorizedAccounts[msg.sender];
+        for (uint i = 0; i < _authorizedAccounts.length; i++) {
+            if (_authorizedAccounts[i].account == authorized) {
+                _authorizedAccounts[i] = _authorizedAccounts[_authorizedAccounts.length - 1];
+                _authorizedAccounts.pop();
+                return;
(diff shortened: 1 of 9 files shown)


contracts/interfaces/payments/IAuthorizedAccount.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Apache-2.0
```
JCSanPedro, line 1: I don't think this would be stored under `interfaces/payments` but just `interfaces/payments`. Reason being is that the delegated signature *may* be used for other purposes in the future, so shouldn't be just scoped to payments.

JCSanPedro (changes requested):
Good job understanding the problem 👍

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,235 @@
...
+            ) {
+                revert AccountDoesNotExist();
+            }
+        }
+    }
+
+    /**
+     * @notice Adds new permissions to a specific authorized account, if the account does not exist, then revert
+     * if any permission in the list has existed in the current permissions, then skip it, otherwise add
+     * new permission at the end of current permissions associated with this authorized account
```
JCSanPedro, line 99: Btw, wording style for `@notice` docs is more of a high level description of what the function will do, as opposed to a detailed description of the algorithm. It should be written as if it was being explained to an end user (not a dev) https://docs.soliditylang.org/en/v0.8.17/natspec-format.html.

So detailing about where exactly the permission will be stored in the array is not very necessary. I would probably write it something like

```
@notice Adds new permissions to a specific authorized account. This will revert if the authorized account does not exist. Adding existing permissions will have no effect.
```

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,235 @@
...
+
+                // replace the old permissions array with the new one.
+                delete _authorizedAccounts[i].permissions;
+                for (uint j = 0; j < counter; j++) {
+                    _authorizedAccounts[i].permissions.push(newPermissions[j]);
+                }
+                break;
+            }
+        }
+    }
```
JCSanPedro, line 186: This function should also fail if the account does not exist

JCSanPedro (changes requested):

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,235 @@
...
+                _authorizedAccounts.pop();
+                break;
+            }
+            if (
+                i == _authorizedAccounts.length - 1 &&
+                _authorizedAccounts[_authorizedAccounts.length - 1].account != authorized
+            ) {
+                revert AccountDoesNotExist();
+            }
+        }
```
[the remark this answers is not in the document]
Helenpham1097: yeah, you are right, I will change

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,219 @@
...
+    /**
+     * @notice Returns true if the contract implements the interface defined by
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(IAuthorizedAccount).interfaceId;
+    }
+
+    /**
+     * @notice Adds new authorized accounts with certain permissions. This will revert if account existed
```
JCSanPedro, line 39:
```suggestion
     * @notice Adds new authorized accounts with certain permissions. This will revert if the account already exists
```

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,219 @@
...
+            account: authorized,
+            createdAt: block.number,
+            permissions: permissions
+        });
+
+        authorizedAccounts[msg.sender].push(newAccount);
+    }
+
+    /**
+     * @notice Removes authorized account of the main account. This will revert if the account
```
JCSanPedro, line 67:
```suggestion
     * @notice Removes an authorized account associated with the msg.sender. This will revert if the account
```

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,219 @@
...
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(IAuthorizedAccount).interfaceId;
+    }
+
+    /**
+     * @notice Adds new authorized accounts with certain permissions. This will revert if account existed
+     * @param authorized The authorized address of the main account
+     * @param permissions The list of certain permissions that the authorized account
```
JCSanPedro, line 41:
```suggestion
     * @param permissions The list of permissions that the authorized account
```

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,219 @@
...
+     * can perform in Sylo network
```
JCSanPedro, line 42:
```suggestion
     * can perform within the Sylo network.
```

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,219 @@
...
+                }
+                return;
+            }
+        }
+        revert AccountDoesNotExist();
+    }
+
+    /**
+     * @notice Removes permissions of specific authorized account. This will revert if the account
+     * does not exist. Removing unexisted permissions will have no effect.
```
JCSanPedro, line 126:
```suggestion
     * does not exist.
```

Probably no need to describe the no-op in this case.

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,219 @@
...
+                }
+                return;
+            }
+        }
+
+        revert AccountDoesNotExist();
+    }
+
+    /**
+     * @notice Validates permission of an authorized account associated with main account
```
JCSanPedro, line 173:
```suggestion
     * @notice Validates permission of an authorized account associated with the main account
```

contracts/AuthorizedAccount.sol
```diff
@@ -0,0 +1,219 @@
...
+                        return true;
+                    }
+                }
+            }
+        }
+        return false;
+    }
+
+    /**
+     * @notice Get all authorized accounts of a given main account
```
JCSanPedro, line 206:
```suggestion
     * @notice Get all authorized accounts associated with the given account
```

JCSanPedro (commented):
Just some minor suggestions regarding doc comments, but otherwise looks good. Will assign to @JoshD641 for review.

JCSanPedro (commented):
Just some minor suggestions regarding doc comments, but otherwise looks good. Will assign to @JoshD641 for review.

JoshD641 (approved):
Looks good to me - the test coverage is thorough, and I can't see any way for the new functionality to be abused in the context of the relay protocol.

JCSanPedro:
@Helenpham1097 Just double checked the coverage and we're not at 100% anymore. Can you add more testing to bump up the coverage to 100%. On that note we should really add a check to the circle ci workflow to fail if 100% isn't reached.

JCSanPedro (approved):

PR #141, opened 2023-06-28
NET-580 Use hardhat-deploy to deploy contracts

teinnt:
hardhat-deploy is a very common hardhat plugin used to manage contract deployments. The plugin has built in support for generating json files that detail the contract deployments. Additionally, using this plugin will allow us to verify contracts on Porcini using sourcify.

This tasks includes upgrading ethers v5 to ethers v6.


hardhat.config.ts
```diff
@@ -1,13 +1,13 @@
 import * as dotenv from 'dotenv';
 
 import { HardhatUserConfig } from 'hardhat/config';
-import '@nomiclabs/hardhat-etherscan';
```
teinnt, line 4: Remove this based on the instruction https://github.com/NomicFoundation/hardhat/releases/tag/%40nomicfoundation%2Fhardhat-toolbox%403.0.0

hardhat.config.ts
```diff
@@ -1,13 +1,13 @@
...
-import '@typechain/hardhat';
-import '@nomiclabs/hardhat-ethers';
+
+import '@nomicfoundation/hardhat-chai-matchers';
+import '@nomicfoundation/hardhat-ethers';
 import 'hardhat-gas-reporter';
+import '@typechain/hardhat';
 import 'solidity-coverage';
-import '@openzeppelin/hardhat-upgrades';
```
teinnt, line 9: Remove this one because we don't use `upgrades` anymore => contract proxy addresses are not saved anymore. It is stated that `hardhat-deploy` saves all deployments scripts [on the disk (git repo)](https://github.com/wighawag/hardhat-deploy) but I don't see it. Should we also saves all deployments by our own?

deploy/001_deploy_contracts.ts
```diff
@@ -0,0 +1,279 @@
...
+): Promise<Receipt> {
+  return await execute(
+    contractName,
+    { from: deployer, log: true },
+    'addManager',
+    manager,
+  );
+}
+
+async function saveContracts(
```
JCSanPedro, line 249: Is there a need for this if `hardhat-deploy` already saves deployments?

deploy/001_deploy_contracts.ts
```diff
@@ -0,0 +1,279 @@
...
+  for (const { name, args } of addManagerParams) {
+    await addManager(name, deployer.address, args[0] as string, execute);
+    printEmptyLine();
+  }
+
+  await saveContracts(deployer.address, network.name, contracts, config);
+};
+
+export default func;
+func.id = 'deploy_contractss'; // id required to prevent reexecution
```
JCSanPedro, line 177:
```suggestion
func.id = 'deploy_contracts'; // id required to prevent reexecution
```

Can you explain the need to do these more? I don't see it in the docs for hardhat-deploy

deploy/001_deploy_contracts.ts
```diff
@@ -0,0 +1,279 @@
...
+    config.SyloToken = (
+      await deployContract('SyloToken', deployer.address, false, deploy)
+    ).address;
+    printEmptyLine();
+  }
+  if (config.Seekers == '') {
+    config.Seekers = (
+      await deployContract('TestSeekers', deployer.address, false, deploy)
+    ).address;
+    printEmptyLine();
```
JCSanPedro, line 63: Seems like you can just add `printEmptyLine` to the `deployContract` fn itself?

deployments/genesis.config.ts
```diff
@@ -63,8 +63,8 @@ const GenesisParameters: ContractParameters = {
 
   TicketingParameters: {
     faceValue: 100000,
-    baseLiveWinProb: BigNumber.from(2).pow(128).sub(1).div(1000),
-    expiredWinProb: BigNumber.from(2).pow(128).sub(1).div(1000),
+    baseLiveWinProb: (BigInt(2) ** BigInt(128) - BigInt(1)) / BigInt(1000),
```
JCSanPedro, line 66: I think you can just do

`baseLiveWinProb: (2n ** 128n - 1n) / 1000n`

hardhat.config.ts
JCSanPedro: Wouldn't it be saved in `./deployments`

JCSanPedro (changes requested):

hardhat.config.ts
teinnt: No, it is not automatically saved in `deployments`. I have downloaded the [example repo](https://github.com/wighawag/template-ethereum-contracts)  and run it but it does not save there.

teinnt: Ah, it didn't save because I run on network hardhat, otherwise it will be saved.

deploy/001_deploy_contracts.ts
teinnt: Actually, I copied this from their template. For further explanation: 
```
The function can return true if it wishes to never be executed again. This can be useful to emulate migration scripts that are meant to be executed only once. Once such script return true (async), the id field is used to track execution and if that field is not present when the script return true, it will fails. In other words, if you want a particular deploy script to run only once, it needs to both return true (async) and have an id set.
```
Since we only want to deploy the factory script once, we should set the `id` field right?

deployments/genesis.config.ts
teinnt: OMG

JCSanPedro (changes requested):
Let's remove the ganache/rata/ropsten deployments since they are not really relevant to us. Otherwise this looks good 👍

teinnt:
Right, I think we should remove `rata` and `ropsten`, however keep `ganache.json` because the `init_network.ts` needs it to connect to contracts, and `ganache.json` is up-to-date with the current testnet deployment.

JCSanPedro (approved):
Good work 👍

PR #142, opened 2023-06-29
NET-590 Add authorized accounts testnet

teinnt:
## Summary of Changes

- Set authorized accounts to testnet nodes in `init_network.ts` to check Sylo Node Testnet later.
- Update Typescript contract names `token` -> `syloToken` and `ticketing` -> `syloTicketing` to be consistent with Solidity contract names.
- Reduce the need to declare/update `SyloContracts` type by declaring the `common` package and put all contract names and types there.
- Remove `rootnet` docker image and setup.


deployments/genesis.config.ts
```diff
@@ -54,7 +54,7 @@ const GenesisParameters: ContractParameters = {
       : '',
 
   EpochsManager: {
-    epochDuration: 80000,
+    epochDuration: 30, // 30 * 4 = 120 seconds = 2 minutes
```
JCSanPedro, line 57: I think for the `GenesisConfig`, this is the config we will use to deploy to mainnet. So it should remain 80000 in this case. We should probably just create another config specifically for the local testnet environment.

JCSanPedro (changes requested):

JCSanPedro (approved):

PR #143, opened 2023-07-25
NET-617 Only update unauthorizedAt if the permission is authorized

teinnt:
So far, if the `_removePermissions` is called, we always update the permission's `unauthorizedAt` field even when it is already unauthorized. User would accidentally create a valid period between the first `unauthorizedAt` and the new `unauthorizedAt`. This PR is to fix the issue and only update `unauthorizedAt` if the permission is authorized.

### DIFF
--- a/contracts/AuthorizedAccounts.sol
+++ b/contracts/AuthorizedAccounts.sol
@@ -205,7 +205,13 @@ contract AuthorizedAccounts is
         for (uint i; i < permissions.length; ++i) {
             for (uint j; j < authAccount.permissions.length; ++j) {
                 if (permissions[i] == authAccount.permissions[j].permission) {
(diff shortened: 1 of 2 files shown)


Nick95550 (approved):
Cool this makes sense to me

JCSanPedro (approved):
Nice 👍

JCSanPedro (changes requested):
Actually, the coverage doesn't show that this if statement isn't covered, but I believe we would need to add a specific test for this?

Unless there just happens to be an existing test that coincidentally covers it?

teinnt:
Yeah, I didn't carefully check, but I remember there are already tests for both cases. I will check it again and let you know.

PR #146, opened 2023-09-06
Net 656 Futurepass Receiver

JCSanPedro:
This PR implements the functionality described here https://www.notion.so/futureverse/RFC-Restricting-broadcasts-to-FP-holders-69119ad113cf4f378360d10e8bf810e7.

The `redeemMultiReceiver` implementation has been modified so that it no longer verifies merkle proofs, but instead ensures that the receiver of the relay has a valid futurepass account.


deployments/genesis.config.ts
```diff
@@ -111,9 +118,14 @@ const GanacheTestnetParameters: ContractParameters = {
...
 
 const PorciniDevParameters: ContractParameters = {
-  SyloToken: '0xCCcCCcCC00000C64000000000000000000000000',
+  // SyloToken: '0xCCcCCcCC00000C64000000000000000000000000',
 
-  Seekers: '0xAAAAAAAA00001864000000000000000000000000',
+  // Seekers: '0xAAAAAAAA00001864000000000000000000000000',
+
+  SyloToken: '',
+  Seekers: '',
```
teinnt, line 126: Should we keep this in the `master` branch?

test/payments/multiReceiverTicketing.test.ts
```diff
@@ -414,36 +405,35 @@ describe('MultiReceiverTicketing', () => {
...
+    const bobs = await Promise.all(
+      Array(5)
+        .fill(0)
+        .map(async _ => {
+          const w = Wallet.createRandom();
+          // register futurepass account
+          await futurepassRegistrar.create(w.address);
+          return w;
+        }),
+    );
```
teinnt, line 417: Should we have a helper function for this?

teinnt (commented):
The change looks great and makes sense to me. 

Just a question, why don't we pass a list of `receivers` and `receiverSigs` to `redeemMultiReceiver`? As calling the method one by one for each receiver will cost more than using a loop, and tickets for later receivers will have less chance to win, since block number will increase.

teinnt:
Btw, this solution is very interesting and reasonable to me. Big brain 😎 🎸🚀

contracts/payments/ticketing/RewardsManager.sol
```diff
@@ -30,6 +30,8 @@ contract RewardsManager is IRewardsManager, Initializable, Manageable, ERC165 {
     // 64x64 Fixed point representation of 1 SYLO (10**18 >> 64)
     int128 internal constant ONE_SYLO_FIXED = 18446744073709551616000000000000000000;
 
+    uint256 internal constant MAX_INT64 = 9223372036854776000;
```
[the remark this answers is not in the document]
JCSanPedro: Ah good find. I don't recall where I got that value from but you're right, it is rounded. I'll update it with the correct value.

teinnt (approved):
LGTM

PR #147, opened 2023-09-19
NET-657 Restrict init first epoch in epochsManager

teinnt:
Epochs are initialized via initializeEpoch which has no restrictions on the caller. This allows any account to initialize epochs as long as the previous duration has elapsed.

This also allows any account to initialize the first epoch and start the network. It would be preferrable to have the method that starts the network be restricted, to ensure that it starts when we want it to.

We can do this by implementing a firstEpoch field which represents the block number the first epoch should start. We can set this field via the constructor and also through a setter setFirstEpoch. The initializeEpoch method will then have another restriction in the case that if epochId == 1, then block.number >= firstEpoch.


contracts/epochs/EpochsManager.sol
```diff
@@ -60,12 +67,14 @@ contract EpochsManager is IEpochsManager, Initializable, Ownable2StepUpgradeable
     error SeekerAccountCannotBeZeroAddress();
     error TicketingParametersCannotBeZeroAddress();
     error SeekerAlreadyJoinedEpoch(uint256 epochId, uint256 seekerId);
+    error TooEarlyToStartFirstEpoch(uint256 expectedBlock, uint256 currentBlock);
```
JCSanPedro, line 70:
```suggestion
    error FirstEpochNotYetReady(uint256 expectedBlock, uint256 currentBlock);
```
I think this follows a pattern where it's like `{variable_name}{error_reason}`
i.e `FirstEpochCannotBeZero` or `SeekerAlreadyJoinedEpoch`

contracts/epochs/EpochsManager.sol
```diff
@@ -227,4 +250,11 @@ contract EpochsManager is IEpochsManager, Initializable, Ownable2StepUpgradeable
     function getNextEpochId() public view returns (uint256) {
         return currentIteration + 1;
     }
+
+    modifier firstEpochMustBeValid() {
```
JCSanPedro, line 254: This modifier isn't being reused anywhere else? Might be more readable to just inline it within the `initializeEpoch` function. That way we don't have to scroll down confirm what it does.

JCSanPedro (changes requested):
Cool looks good. Reading through the code I think it "feels" better to call it `InitialEpoch` instead of `FirstEpoch`. We tend to use the descriptor `initial` throughout the code base more.

JCSanPedro (approved):
Fantastic :)

PR #160, opened 2024-01-15
SM-117 Seeker Oracle

JCSanPedro:
This PR adds the interface for the Seeker Oracle contract, as well as a basic implementation. The Seeker Oracle works by keeping a record of a "oracle" account, and also a mapping of Seeker Ids to Seeker Power levels. The oracle account to set the a Seeker's power level mapping by calling the method `setSeekerPowerRestricted`. Only the oracle account and the owner account (deployer) are allowed to call this method.

There is also an additional way to set the seeker's power level by calling the unrestricted method `registerSeekerPower`. The register method takes an additional parameter, a `proof` message. The proof is a specific message that has been signed by the `oracle` account. The structure of the message is as follows:
`${seekerId}:${seekerPower}`
The register function only allows proofs that have been signed by the `oracle` account.

This approach allows us to build an API that holds the `oracle` account private key, and provides an API method to acquire the proof for a given Seeker Id. The API has knowledge of all Seeker power levels (and acts as the source of truth outside of the contracts).

### DIFF
--- a/contracts/SeekerPowerOracle.sol
+++ b/contracts/SeekerPowerOracle.sol
@@ -49,7 +49,7 @@ contract SeekerPowerOracle is
   }
 
   function setSeekerPowerRestricted(uint256 seekerId, uint256 power) external {
-    if (msg.sender != this.owner() || msg.sender != oracle) {
+    if (msg.sender != this.owner() && msg.sender != oracle) {
       revert UnauthorizedSetSeekerCall();
     }
 
@@ -57,8 +57,8 @@ contract SeekerPowerOracle is
     emit SeekerPowerUpdated(seekerId, power);
   }
 
-  function setSeekerPower(uint256 seekerId, uint256 power, bytes calldata proof) external {
-    bytes memory proofMessage = getProofMessage((seekerId, power);
+  function registerSeekerPower(uint256 seekerId, uint256 power, bytes calldata proof) external {
+    bytes memory proofMessage = getProofMessage(seekerId, power);
     bytes32 ecdsaHash = ECDSA.toEthSignedMessageHash(proofMessage);
 
     if (ECDSA.recover(ecdsaHash, proof) != oracle) {
@@ -73,7 +73,7 @@ contract SeekerPowerOracle is
     return seekerPowers[seekerId];
   }
 
-  function getProofMessage(uint256 seekerId, uint256 power) external view returns (bytes memory) {
+  function getProofMessage(uint256 seekerId, uint256 power) public pure returns (bytes memory) {
     return abi.encodePacked(
       Strings.toString(seekerId),
       ":",
--- a/contracts/SeekerPowerOracle.sol
+++ b/contracts/SeekerPowerOracle.sol
@@ -13,71 +13,112 @@ import "./interfaces/ISeekerPowerOracle.sol";
  * a Seeker's power level via a restricted owner call. Seeker Power can also
  * be set by any account if the correct Oracle signature proof is provided.
  */
-contract SeekerPowerOracle is
-  ISeekerPowerOracle,
-  Initializable,
-  Ownable2StepUpgradeable
-{
-  /**
-   * @notice The oracle account. This contract accepts any attestations of
-   * Seeker power that have been signed by this account.
-   */
-  address public oracle;
+contract SeekerPowerOracle is ISeekerPowerOracle, Initializable, Ownable2StepUpgradeable {
+    /**
+     * @notice The oracle account. This contract accepts any attestations of
+     * Seeker power that have been signed by this account.
+     */
+    address public oracle;
 
-  /**
-   * @notice Tracks the set Seeker Power levels.
-   */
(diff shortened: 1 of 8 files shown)


contracts/SeekerPowerOracle.sol
```diff
@@ -0,0 +1,124 @@
...
+import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
+import "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
+import "@openzeppelin/contracts/utils/cryptography/ECDSA.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
+
+import "./interfaces/ISeekerPowerOracle.sol";
+
+/**
+ * @notice Acts as a source of information for Seeker Powers. Allows setting
+ * a Seeker's power level via a restricted owner call. Seeker Power can also
```
teinnt, line 13: Should we say `via a restricted owner or oracle call` to be clear?

contracts/SeekerPowerOracle.sol
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * @notice Registers a Seeker's power level. Only callable by the
+     * owner or the oracle account.
+     * @param seekerId The id of the Seeker.
+     * @param power The power level of the Seeker.
+     */
+    function registerSeekerPowerRestricted(uint256 seekerId, uint256 power) external {
+        if (msg.sender != this.owner() && msg.sender != oracle) {
```
teinnt, line 60: Hmm, I think only the `oracle` account has the right to register seeker power. This way would be more clear and strict. What is the reason to allow `owner` to do that?

contracts/SeekerPowerOracle.sol
```diff
@@ -0,0 +1,124 @@
...
+
+    event SeekerPowerUpdated(uint256 indexed seekerId, uint256 indexed power);
+
+    function initialize(address _oracle) external initializer {
+        Ownable2StepUpgradeable.__Ownable2Step_init();
+
+        oracle = _oracle;
+    }
+
+    error UnauthorizedSetSeekerCall();
```
teinnt, line 42: Should we name it something like `UnauthorizedSenderToRegisterSeekerPower`? `SetSeeker` can be confused with the `Registries` contract.

```suggestion
    error UnauthorizedSetSeekerCall();
```

contracts/SeekerPowerOracle.sol
```diff
@@ -0,0 +1,124 @@
...
+        bytes32 ecdsaHash = ECDSA.toEthSignedMessageHash(proofMessage);
+
+        if (ECDSA.recover(ecdsaHash, proof) != oracle) {
+            revert UnauthorizedSetSeekerCall();
+        }
+
+        seekerPowers[seekerId] = power;
+        emit SeekerPowerUpdated(seekerId, power);
+
+        proofNonces[nonce] = oracle;
```
teinnt, line 94: Should we update this variable before `seekerPowers`, so that we can emit event at the end of the method? I remember we should always emit event after everything, but you can decide.

contracts/SeekerPowerOracle.sol
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    /**
+     * @notice Retrieves a Seeker's stored power level.
+     * @param seekerId The id of the Seeker.
+     */
+    function getSeekerPower(uint256 seekerId) external view returns (uint256) {
+        return seekerPowers[seekerId];
+    }
```
teinnt, line 103: I feel it may be confused, like, do we have a rule like: Seeker cannot have 0 power? Otherwise we don't know if this seeker is not set or have 0 power. Unless we add another restriction `CannotSetZeroPower`.

test/utils.ts
```diff
@@ -37,6 +38,8 @@ const initializeContracts = async function (
 
   const minimumStakeProportion = opts.minimumStakeProportion ?? 2000;
 
+  const oracle = opts.oracle ?? deployer;
```
teinnt, line 41: Should we name `seekerPowerOracleAddr` to be clear? For both constant and options's name.


Or can we just do as below in line 129?
```suggestion
  await seekerPowerOracle.initialize(opts.oracle ?? deployer, { from: deployer });
```

contracts/SeekerPowerOracle.sol
JCSanPedro: Yeah agreed. I did it to make testing easier for us but we can just set the deployer/oracle to the same account anyway

contracts/SeekerPowerOracle.sol
JCSanPedro: A seeker will have more than 0 power. So we can safely assume that 0 power just means not registered.

test/utils.ts
JCSanPedro: I put the `const oracle` to keep in line with how we defined the other default options. I think we can renames this to `seekerPowerOracle` just to be clearer.

teinnt (approved):
Perfect!!!

PR #161, opened 2024-01-17
SM-118 Staking Power

JCSanPedro:
The changes in this PR include using the SeekerPowerOracle to determine a Seeker's staking capacity. It introduces a new method to the staking manager `calculateCapacityFromSeekerPower` that first queries a Seeker's power level then uses that value to determine it's staking capacity. The current algorithm is as follows:
```
staking_capacity = seeker_power ** seeker_power_multiplier
```

Seeker power multipler is a constant that we can change through a setter. This will give us flexibility post deployment.

The Directory contract has been updated to call `calculateCapacityFromSeekerPower` when a node joins an epoch. The maximum amount of stake the node can join an epoch with will also be determined by the staking capacity value. The maximum joined stake cannot exceed either seeker stake cap, or the minimum stake proportion stake. However for the next release, we will set the min staking proportion value to a very low number (1/10000). This requires a change in the percentage denominator to go from a uint16 value to a uint32 value.

The deployment script has also been updated to support deploying the SeekerPowerOracle contract.


teinnt (commented):
Awesome, LGTM!

teinnt:
I don't see any deployment updates and ganache contracts update in this PR. Is it missing the last commit?

test/staking.test.ts
```diff
@@ -994,6 +1009,135 @@ describe('Staking', () => {
...
+
+      expect(capacity).to.equal(expectedSylo);
+    }
+  });
+
+  it('returns maximum SYLO amount if seeker power is very large', async () => {
+    const maxSylo = ethers.parseEther('10000000000');
+
+    // 1 more than the maximum sylo
+    await seekerPowerOracle.registerSeekerPowerRestricted(111, maxSylo + 1n);
```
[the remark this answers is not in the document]
JCSanPedro: This is also tested because we want to avoid multiplying the power if it would cause an overflow

test/staking.test.ts
```diff
@@ -994,6 +1015,168 @@ describe('Staking', () => {
...
+    // the added stake is 10,000,000 SYLO, but the seeker power capacity
+    // is 4,000,000
+    await seekerPowerOracle.registerSeekerPowerRestricted(111, 4);
+
+    await directory.addManager(owner);
+    await directory.joinNextDirectory(owner, 111);
+
+    const joinedStake = await directory.getTotalStakeForStakee(1, owner);
+
+    expect(joinedStake).to.equal(ethers.parseEther('4000000'));
```
teinnt, line 1146: I was thinking, should we have an event when the staking amount is deducted, so that we or user can query it when needed?

teinnt (approved):
Lovely!

contracts/staking/Directory.sol
```diff
@@ -112,22 +126,23 @@ contract Directory is IDirectory, Initializable, Manageable, IERC165 {
...
 
         uint256 joiningStake;
-        if (ownedStakeProportion >= minimumStakeProportion) {
+        if (totalStake > seekerStakingCapacity && totalStake > minProportionStakingCapacity) {
+            joiningStake = Math.min(seekerStakingCapacity, minProportionStakingCapacity);
+        } else if (totalStake > seekerStakingCapacity) {
+            joiningStake = seekerStakingCapacity;
+        } else if (totalStake > minProportionStakingCapacity) {
+            joiningStake = minProportionStakingCapacity;
+        } else { // uncapped
```
teinnt, line 142: Can we use switch case?

deploy/001_deploy_contracts.ts
```diff
@@ -167,8 +177,6 @@ function getConfig(networkName: string): configs.ContractParameters {
   switch (networkName) {
     case 'porcini-dev':
       return configs.PorciniDevParameters;
-    case 'porcini-testing':
-      return configs.PorciniTestingParameters;
```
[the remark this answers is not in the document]
JCSanPedro: We just haven't been using it. Adds confusion if you select the wrong one.

deploy/001_deploy_contracts.ts
```diff
@@ -130,6 +136,10 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
         config.Ticketing.unlockDuration,
       ],
     },
+    {
+      name: ContractNames.seekerPowerOracle,
+      args: [config.SeekerPowerOracle.oracleAccount],
+    },
```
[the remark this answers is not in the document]
JCSanPedro: It is one of our deployed contracts

contracts/staking/Directory.sol
JCSanPedro: I don't think there is a switch case in solidity?

test/staking.test.ts
JCSanPedro: It's already possible to see the amount of stake the node joined the directory with

teinnt (approved):

PR #162, opened 2024-02-22
SM-214 Contracts

JCSanPedro:
Deploys a new set of contracts for testing, and adjusts the parameters again.

We also commit the deployment data (`deployments/porcini-dev/*`), as we need these files to perform upgrades post deployment.


Nick95550 (approved):
LGTM

PR #163, opened 2024-03-05
SM-251 Mainnet

JCSanPedro:
This PR focuses on saving a mainnet deployment of the protocol contracts.

We also add a fix to the staking manager contract to have a function to set the seeker power multiplier. This fix required a contract upgrade as I had already deployed a set of contracts to mainnet.


Nick95550 (approved):
LGTM

PR #164, opened 2024-03-10
SM-284 Fix Last Claim edge case

JCSanPedro:
The rewards contracts uses a specific technique to optimize the gas cost of claiming rewards over multiple epochs. Traditionally, for each epoch we would need to track each staker's individual stake contribution, and then calculate the staker's reward share for each epoch. This significantly increases the gas cost for the node operator when their node joins an epoch (the transaction would need to perform a write call for each of the node's stakers), and also means that the claim rewards transaction will linearly scale with number of epochs the claim is being made for. The gas cost would become excessive if the user had not claimed for a long period.

To avoid this issue, we utilize a variable called the cumulative reward factor https://github.com/futureversecom/sylo-ethereum-contracts/blob/master/docs/spec.md#reward-calculation-and-cumulative-reward-factor.

The problem with this approach is that it assumes the staker's stake is fixed for the period of the calculation. This means that whenever the stake changes (or a claim is made), we need to perform the reward share calculation for that staker, and move any pending rewards to an unclaimed reward variable. 

```
Example

1. User has 1000 stake at the start of the network.
2. User adds 1000 more stake at epoch 10 (total 2000 stake)
    a. Contract calculates rewards from epoch 1 to 9, with users stake at 1000, and moves reward from pending to unclaimed.
3. User add 1000 more stake at epoch 20
    a. Contract calculates rewards from epoch 11 to 19, with users stake at 2000, and moves reward from pending to unclaimed.
```

In the above example we can see that epoch `10` was missed. This is because when a claim is made, we only calculate up to the end of the previous epoch to ensure there are no rewards missed.  To account for this missed epoch, we store and update a `LastClaim` variable for that staker whenever a reward calculation is performed.

```
LastClaim = {
  uint256 claimedAt; // the epoch the claim was performed
  uint256 stake; // The amount of stake the user had when this claim was made
}
```

When a `lastClaim` variable is updated, it describes the epoch the update was made, and the amount of stake the user has for that epoch. The calculation is then updated to use the last claim variable to account for the first epoch.

```
Example

1. User has 1000 stake at the start of the network.
2. User adds 1000 more stake at epoch 10 (total 2000 stake)
    a. Contract calculates rewards from epoch 1 to 9, with users stake at 1000, and moves reward from pending to unclaimed.
    b. Contract stores last claim = { claimAt: 10, stake: 1000 }
3. User claims rewards at epoch 20
    a. Contract calculates initial claim using the last claim stored for that user (epoch 10 with stake 1000) 
    b. Contract calculates rewards from epoch 11 to 19, with users stake at 2000
    c. Contract sums calcs from step a and b together as the final claim
    d. Contract updates last claim = { claimAt: 20, stake: 2000 }
```

However the contract implementation did not account for the edge case when a user stakes or claims once the node has already joined the next epoch. When a node joins the next epoch, it locks in its stake for that epoch. Any changes to stake from that point on will be applied to the epoch after the next one. The way we updated the last claim variable did not account for this.

```
Example
1. User adds 1000 more stake at epoch 10 (total 2000 stake)
    a. Contract calculates rewards from epoch 1 to 9, with users stake at 1000, and moves reward from pending to unclaimed.
    b. Contract stores last claim = { claimAt: 10, stake: 1000 }
2. Node joins epoch and locks in stake
3. Before next epoch starts, user adds 1000 more stake
    a. Contract will calculate claim but the calculation will return 0 as last claim is at 10 already. However the last claim variable gets updated to = { claimAt: 10, stake: 2000 }
4. User claims rewards at epoch 20
    a. Contract calculates initial claim using the last claim stored for that user (epoch 10 with stake 2000) <-- XXX This is where the calculation goes wrong. The actual stake it is using for the initial claim should be 1000 not 2000. 
```

At step `4a`, the initial claim will be off and the reward amount will be off as well. This can eventually lead to an underflow over time.

## Fix

To prevent this edge case from occurring, we revert if a claim/stake change is made once the node has already initialized the next reward pool.

## Simulation

We also implement a simulation test. It does not run unless the env var `RUN_SIMULATIONS=1` is set. The simulation initializes 10 nodes and 25 stakers, and simulates various stake actions occurring over several epochs. At the end of the loop, all stakers should be able to make a claim without any reverts occurring.


contracts/payments/ticketing/RewardsManager.sol
```diff
@@ -541,6 +542,11 @@ contract RewardsManager is IRewardsManager, Initializable, Manageable, ERC165 {
             revert StakeeCannotBeZeroAddress();
         }
 
+        uint256 currentEpoch = _epochsManager.currentIteration();
+        if (getRewardPool(currentEpoch + 1, stakee).totalActiveStake > 0) {
+            revert NextRewardPoolAlreadyActive();
+        }
```
teinnt, line 548: As this repeats in other places, should we use `modifier`?

contracts/payments/ticketing/RewardsManager.sol
```diff
@@ -576,13 +587,23 @@ contract RewardsManager is IRewardsManager, Initializable, Manageable, ERC165 {
     }
 
     function updateLastClaim(address stakee, address staker) internal {
+        bytes32 stakerKey = getStakerKey(stakee, staker);
+        LastClaim storage lastClaim = lastClaims[stakerKey];
+
+        uint256 claimAt = _epochsManager.currentIteration();
```
teinnt, line 593: Why do we move those lines up here? In case the current epoch is updated in another transaction?

test/simulations.test.ts
```diff
@@ -0,0 +1,303 @@
...
+    // have each staker randomly stake to 5 nodes
+    await Promise.all(
+      stakers.map(async staker => {
+        stakeEntries[staker.address] = {};
+
+        for (let i = 0; i < 5; i++) {
+          // select random node
+          const node = nodes[randomInt(nodeCount)];
+
+          // stake between 1 and 1000000
```
teinnt, line 94:
```suggestion
          // stake between 1 and 1000001
```

test/simulations.test.ts
```diff
@@ -0,0 +1,303 @@
...
+        await contracts.stakingManager.getAddress(),
+        ethers.parseEther('10000000000'),
+      );
+
+    return { address, signer };
+  };
+
+  // for each of a node's stakers, run an action that will
+  // affect their stake
+  const runRandomStakerAction = async (
```
teinnt, line 233: Whooooo

test/utils.ts
```diff
@@ -186,11 +187,72 @@ const initializeContracts = async function (
   };
 };
 
+const onlyDeployContracts = async function (
```
teinnt, line 190: Do we use it anywhere?

teinnt (commented):
That is amazing that you found the root cause and the explanation is very clear. I believe this should be kept in README later.

contracts/payments/ticketing/RewardsManager.sol
JCSanPedro: Felt the need to refactor it this way to make the behaviour more clear.

test/simulations.test.ts
JCSanPedro: It's actually 1 and 1000000 as the `randomInt(x)` returns a value from `0 to (x - 1)`

test/utils.ts
JCSanPedro: No but I found it useful when writing other simulations so want to keep it in just for now.

test/simulations.test.ts
teinnt: Alright, it rounds down.

teinnt (approved):

PR #168, opened 2024-05-01
SM-329 Attached Authorized Account

JCSanPedro:
This PR adds an additional option for authorized accounts where the node can supply the authorized account for a receiver when redeeming tickets.

A new struct is introduced `AttachedAuthorizedAccount`. This struct contains the following:
- `account` The address of the authorized account
- `expiry` A unix timestamp in seconds to indicate when this account is no longer valid
- `proof` A message signed by the user's main account to act as proof of authorization
- `prefix`, `suffix`, and `infix` are strings used to construct the proof message. We use these values to allow experiences to create their own human-friendly proof messages.

The `IAuthorizedAccounts` also added a method for creating a proof message, and validating an attached authorized account.

To utilize the above solution, two new methods were also added to the ticketing contract: `redeemV2` and `redeemMultiReceiverV2`. These methods act very similar to the existing redeem methods, though allow specifying an authorized account for the signatures in the parameters. V2 version were made as changing the parameters is a breaking change.

The V2 will additionally validate the attached authorized accounts if they are present.


contracts/payments/SyloTicketing.sol
```diff
@@ -321,6 +322,51 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
...
+     * DEPRECATED in favour of `redeemV2`.
+     */
+    function redeem(
+        Ticket calldata ticket,
+        uint256 redeemerRand,
+        bytes calldata _senderSig,
+        bytes calldata _receiverSig
+    ) external {
+        SignatureType senderSigType;
+        SignatureType receiverSigType;
```
teinnt, line 335: Should we declare `UserSignature memory senderSig` and `receiverSig` here, so we dont need to use extra memory for `senderSigType` and `receiverSigType`?

Also, I've just tested and the default value of `sigType` is `Main`, so we just need to assign value if the type is `Authorized`.

```
        UserSignature memory senderSig;
        UserSignature memory receiverSig;

        if (ticket.sender.delegated != address(0)) {
            senderSig.sigType = SignatureType.Authorized;
        }
        if (ticket.receiver.delegated != address(0)) {
            receiverSig.sigType = SignatureType.Authorized;
        }

        senderSig.signature = _senderSig;
        receiverSig.signature = _receiverSig;

        (EpochsManager.Epoch memory epoch, bytes32 ticketHash) = validateRedeem(
            ticket,
            redeemerRand,
            senderSig,
            receiverSig
        );
```

contracts/payments/SyloTicketing.sol
```diff
@@ -515,24 +646,41 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
...
+        if (
+            receiverSig.sigType == SignatureType.Authorized &&
+            !hasSigningPermission(ticket.receiver, ticket.generationBlock)
+        ) {
             revert InvalidReceiverSigningPermission();
         }
 
-        if (!isValidTicketSig(ticket.sender, senderSig, ticketHash)) {
+        if (senderSig.sigType == SignatureType.AttachedAuthorized) {
+            revert SenderCannotUseAttachedAuthorizedAccount();
```
teinnt, line 663: Maybe we add a comment in the `redeemV2` and `redeemMultiReceiverV2` that sender cannot use the attached authorized account?

contracts/payments/SyloTicketing.sol
```diff
@@ -515,24 +646,41 @@ contract SyloTicketing is ISyloTicketing, Initializable, Ownable2StepUpgradeable
             revert RedeemerCommitMismatch();
         }
 
-        if (!hasSigningPermission(ticket.sender, ticket.generationBlock)) {
```
teinnt, line 518: Hmm, to be honest I think we can make if-else to reduce extra check and extra fee. But the code is pretty ugly though. Also `requireValidWinningMultiReceiverTicket`.

```
        // validate sender signature
        if (
            senderSig.sigType == SignatureType.Authorized &&
            !hasSigningPermission(ticket.sender, ticket.generationBlock)
        ) {
            revert InvalidSenderSigningPermission();
        } else if (senderSig.sigType == SignatureType.AttachedAuthorized) {
            revert SenderCannotUseAttachedAuthorizedAccount();
        } else if (!isValidTicketSig(ticket.sender, senderSig.signature, ticketHash)) {
            revert InvalidSenderSignature();
        }

        // validate receiver signature
        if (
            receiverSig.sigType == SignatureType.Authorized &&
            !hasSigningPermission(ticket.receiver, ticket.generationBlock)
        ) {
            revert InvalidReceiverSigningPermission();
        } else if (receiverSig.sigType == SignatureType.AttachedAuthorized) {
            _authorizedAccounts.validateAttachedAuthorizedAccount(
                ticket.receiver.main,
                receiverSig.attachedAuthorizedAccount
            );
        } else if (!isValidTicketSig(ticket.receiver, receiverSig.signature, ticketHash)) {
            revert InvalidReceiverSignature();
        }
```

test/authorizedAccount.test.ts
```diff
@@ -816,6 +816,204 @@ describe('Authorized Accounts', () => {
...
+          infixOne,
+        },
+      ),
+    ).to.be.revertedWithCustomError(
+      authAccountsConnectMain,
+      'AttachedAuthorizedAccountExpired',
+    );
+  });
+
+  it('reverts with error is authorized account is not signed by main account', async () => {
```
teinnt, line 899: Can we add the case that the delegated account is changed?

test/payments/multiReceiverTicketing.test.ts
```diff
@@ -516,6 +521,123 @@ describe('MultiReceiverTicketing', () => {
...
+    );
+
+    await syloTicketing.redeemMultiReceiverV2(
+      ticket,
+      redeemerRand,
+      { main: bob.address, delegated: delegatedWallet.address },
+      {
+        sigType: SignatureType.Main,
+        signature: senderSig,
+        attachedAuthorizedAccount: createEmptyAttachedAuthorizedAccount(),
```
teinnt, line 570: It is quite annoying, but totally agree that we pass the same struct for sender and receiver.

test/payments/utils.ts
```diff
@@ -282,3 +283,50 @@ export function createCommit(
...
 }
+
+export async function createAttachedAuthorizedAccount(
+  main: HDNodeWallet,
+  delegatedWallet: HDNodeWallet,
+  authorizedAccounts: contractTypes.AuthorizedAccounts,
+): Promise<IAuthorizedAccounts.AttachedAuthorizedAccountStruct> {
+  const block = await ethers.provider.getBlock('latest');
+
+  const expiry = (block?.timestamp ?? 0) + 10000000;
```
teinnt, line 294: Should we make `10000000` a constant if we use timestamp?

teinnt (commented):
Wow, a lot of changes! All redeem and redeemV2 functions looking all great!!! Amazing PR John!!!!

contracts/AuthorizedAccounts.sol
```diff
@@ -302,4 +306,72 @@ contract AuthorizedAccounts is
...
+                Strings.toString((expiry)),
+                suffix
+            );
+    }
+
+    function validateAttachedAuthorizedAccount(
+        address main,
+        AttachedAuthorizedAccount calldata account
+    ) external view {
+        if (block.timestamp > account.expiry) {
```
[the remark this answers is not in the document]
JCSanPedro: I think we will try to transition towards timestamps instead of block numbers (e.g using timestamp for epoch duration) for the next seeker staking phase, so wanted to start here.

I think timestamps are okay for us because the types of attacks where ethereum nodes can manipulate timestamps doesn't really apply to us.

test/payments/utils.ts
JCSanPedro: I don't think it is necessary to define constants just for unit tests. In this case the value isn't really meaningful, it just has to represent some time in the future.

contracts/AuthorizedAccounts.sol
```diff
@@ -302,4 +306,72 @@ contract AuthorizedAccounts is
...
+    /**
+     * @notice Creates a proof for an authorized account. The prefix, suffix,
+     * and infix strings allow applications to create more human-readable
+     * messages for personal_sign requests.
+     * @param account The authorized account
+     * @param prefix String that is prefixed before the account address in the
+     * signing message
+     * @param suffix String that is placed after the expiry value in the signing
+     * message
+     * @param infixOne String that is placed between the address and expiry in
```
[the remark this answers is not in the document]
JCSanPedro: Having `prefix`, suffix`, and `infixOne` will other experience developers more flexibility when constructing the personal message to sign. It would be even possible to define default values for this if other devs/designers don't want to think of the copy, but I don't think that is even necessary.

teinnt (approved):
LGTM!!!

test/payments/multiReceiverTicketing.test.ts
```diff
@@ -370,9 +389,9 @@ describe('MultiReceiverTicketing', () => {
           generationBlock: BigInt(ticket.generationBlock) + 10n,
         },
         redeemerRand,
-        { main: bobs[0].address, delegated: ethers.ZeroAddress },
+        bobs[0].address,
+        senderSig,
         senderSig,
```
Nick95550, line 394:
```suggestion
        bobs[0].address,
        senderSig,
        new Uint8Array(0),
```
up to you

test/payments/multiReceiverTicketing.test.ts
```diff
@@ -509,16 +538,113 @@ describe('MultiReceiverTicketing', () => {
...
+      authorizedAccounts,
+    );
+
+    await syloTicketing.redeemMultiReceiver(
+      ticket,
+      redeemerRand,
+      bob.address,
+      senderSig,
+      receiverSig,
+    );
```
Nick95550, line 591:
```suggestion
    const receiverSig = await createUserSignature(
      ethers.getBytes(ticketHash),
      SignatureType.AttachedAuthorized,
      bob,
      delegatedWallet,
      authorizedAccounts,
    );

    console.log('receiver sig ', receiverSig);

    await syloTicketing.redeemMultiReceiver(
      ticket,
      redeemerRand,
      bob.address,
      senderSig,
      receiverSig,
    );
```
I think this is supposed to be attactched ?

Nick95550 (changes requested):
Sweet as, just a couple minor comments

test/payments/multiReceiverTicketing.test.ts
JCSanPedro: The isn't just a Uint8Array but is actually a struct, which is why I'm reusing it here.

test/payments/ticketing.test.ts
```diff
@@ -1193,6 +1198,115 @@ describe('Ticketing', () => {
...
+      to: alice.address,
+      value: ethers.parseEther('2000.0'),
+    });
+
+    const attachedAuthorizedAccount = await createAttachedAuthorizedAccount(
+      alice,
+      delegatedWallet,
+      authorizedAccounts,
+    );
+
```
Nick95550, line 1287: I think this is still unused

Nick95550 (changes requested):
Just two minors

Nick95550 (approved):
LGTM

PR #171, opened 2024-06-05
Sm 350 seekers stats oracle

Nick95550:
Implementes the interface and contract for Seeker Stats Oracle


contracts/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,205 @@
+// SPDX-License-Identifier: Apache-2.0
```
JCSanPedro, line 1: I think we can put this file under
```
contracts/
  staking/
    seekers/
      SeekerStatsOracle.sol
```

contracts/ISeekerStatsOracle.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.18;
+
+interface ISeekerStatsOracle {
+    struct Seeker {
+        uint256 seekerId;
+        uint256 rank;
+        uint256 attr_chip;
```
JCSanPedro, line 8: Lets name the attribute fields in mixedCase
```
uint256 attrChip;
uint256 attrDurability;
...
```

contracts/ISeekerStatsOracle.sol
```diff
@@ -0,0 +1,27 @@
...
+        uint256 attr_reactor;
+    }
+
+    function setOracle(address _seekerStatsOracleAccount) external;
+
+    function createProofMessage(Seeker calldata seeker) external pure returns (bytes memory);
+
+    function registerSeekerRestricted(Seeker calldata seeker) external;
+
+    function registerSeeker(Seeker calldata seeker, bytes calldata signature) external;
```
JCSanPedro, line 22:
```suggestion
    function registerSeeker(Seeker calldata seeker, bytes calldata proof) external;
```

contracts/ISeekerStatsOracle.sol
```diff
@@ -0,0 +1,27 @@
...
+
+    function calculateAttributeCoverage(
+        Seeker[] calldata seekersList
```
JCSanPedro, line 25:
```suggestion
        Seeker[] calldata seekers
```

contracts/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    /**
+     * @notice Holds the angle used for coverage calculation in radians
+     */
+    int256 private coverageAngle =
+        Trigonometry.sin(((Trigonometry.TWO_PI / 6) + Trigonometry.TWO_PI));
+
+    event SeekerStatsUpdated(
+        uint256 indexed seekerId,
+        uint256 attr_reactor,
```
JCSanPedro, line 32:
```suggestion
        uint256 attrReactor,
```

Let's use mixedCase since that is the expected casing for solidity

contracts/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,205 @@
...
+        uint256 attr_cores,
+        uint256 attr_durability,
+        uint256 attr_sensors,
+        uint256 attr_storage,
+        uint256 attr_chip
+    );
+
+    error OracleCannotBeZeroAddress();
```
JCSanPedro, line 40:
```suggestion
    error OracleAddressCannotBeNil();
```

contracts/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,205 @@
...
+import "solidity-trigonometry/src/Trigonometry.sol";
+
+import "./ISeekerStatsOracle.sol";
+
+contract SeekerStatsOracle is ISeekerStatsOracle, Initializable, Ownable2StepUpgradeable, ERC165 {
+    /**
+     * @notice The oracle account. This contract accepts any attestations of
+     * Seeker power that have been signed by this account.
+     */
+    address public SeekerStatsOracleAccount;
```
JCSanPedro, line 17:
```suggestion
    address public oracle;
```

contracts/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,205 @@
...
+     * @param seeker The object containing the seekers statistics.
+     * @param signature The signature of the seekers proof message, signed by the oracle account.
+     */
+    function validateSeekerStatsProof(
+        Seeker calldata seeker,
+        bytes calldata signature
+    ) internal view returns (bool) {
+        if (SeekerStatsOracleAccount == address(0)) {
+            revert OracleCannotBeZeroAddress();
+        }
```
JCSanPedro, line 88: Not really necessary here since we already have this requirement whenever we set the oracle account

contracts/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,205 @@
...
+            sumCoverage +=
+                (int256(seeker.attr_sensors) * coverageAngle * int256(seeker.attr_storage)) /
+                2;
+            sumCoverage +=
+                (int256(seeker.attr_storage) * coverageAngle * int256(seeker.attr_chip)) /
+                2;
+            sumCoverage +=
+                (int256(seeker.attr_chip) * coverageAngle * int256(seeker.attr_reactor)) /
+                2;
+        }
```
JCSanPedro, line 201: I think this implementation is not quite right. Seems like you're calculating the coverage for each individual Seeker, then summing them all up.

What you need to is first sum up all of the attributes, then calculate the coverage for the summed attributes.

```
uint256 totalReactor = 0;
uint256 totalCores = 0;
...

for (...) {
  totalReactor += seeker.attrReactor;
  totalCores += seeker.attrCores;
  ...
}

// calculate coverage from the summed attributes
```

JCSanPedro (changes requested):
Great work 👍 Just some minors

contracts/staking/seekers/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,202 @@
...
+            totalStorage += int256(registeredSeeker.attrStorage);
+            totalChip += int256(registeredSeeker.attrChip);
+        }
+
+        coverage += (int256(totalReactor) * coverageAngle * int256(totalCores)) / 2;
+        coverage += (int256(totalCores) * coverageAngle * int256(totalDurability)) / 2;
+        coverage += (int256(totalDurability) * coverageAngle * int256(totalSensors)) / 2;
+        coverage += (int256(totalSensors) * coverageAngle * int256(totalStorage)) / 2;
+        coverage += (int256(totalStorage) * coverageAngle * int256(totalChip)) / 2;
+        coverage += (int256(totalChip) * coverageAngle * int256(totalReactor)) / 2;
```
teinnt, line 198: Good that is is a view function, so we don't need to consider gas cost of the calculation. Cool 🚀🚀🚀

contracts/staking/sylo/ISyloStakingManager.sol
```diff
@@ -0,0 +1,54 @@
...
+
+interface ISyloStakingManager {
+    /**
+     * For every Node, there will be a mapping of the staker to a
+     * StakeEntry. The stake entry tracks the amount of stake in SOLO,
+     * and also when the stake was updated.
+     */
+    struct StakeEntry {
+        uint256 amount;
+        // Timestamp this entry was updated at
```
teinnt, line 12:
```suggestion
        // Timestamp this entry was updated at: block.timestamp
```

contracts/staking/sylo/ISyloStakingManager.sol
```diff
@@ -0,0 +1,54 @@
...
+        uint256 updatedAt;
+    }
+
+    /**
+     * Every Node must have stake in order to participate in the Epoch.
+     * Stake can be provided by the Node itself or by other accounts in
+     * the network.
+     */
+    struct Stake {
+        // Track each stake entry associated to a node
```
teinnt, line 22:
```suggestion
        // Tracks each stake entry associated to a node
```

contracts/staking/sylo/ISyloStakingManager.sol
```diff
@@ -0,0 +1,54 @@
...
+        mapping(address => StakeEntry) entries;
+        // The total stake held by this contract for a node,
+        // which will be the sum of all addStake and unlockStake calls
```
teinnt, line 25: Idk, I guess this one is a bit easier to remmeber.

```suggestion
        // The sum of all addStake and unlockStake calls, held by this contract for one node
```

contracts/staking/sylo/ISyloStakingManager.sol
```diff
@@ -0,0 +1,54 @@
...
+        uint256 totalManagedStake;
+    }
+
+    /**
+     * This struct will track stake that is in the process of unlocking.
+     */
```
teinnt, line 31: I think if we want to comment, we can put more information or explanation. Otherwise I think `Unlocking` is clear enough to not have the comment above. You can decide.

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+import "./ISyloStakingManager.sol";
+
+contract SyloStakingManager is
+    ISyloStakingManager,
+    Initializable,
+    Ownable2StepUpgradeable,
+    ERC165
+{
+    /** IERC20 Sylo Token address */
+    IERC20 public _sylo;
```
teinnt, line 18: Maybe name it `syloToken` to be safe and clear?

```suggestion
    IERC20 public _syloToken;
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * @notice Tracks the managed stake for every Node.
+     */
+    mapping(address => Stake) public stakes;
+
+    /** @notice Tracks overall total stake held by this contract */
```
teinnt, line 25:
```suggestion
    /** 
     * @notice Tracks overall total stake held by this contract 
     */
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * @notice The duration in seconds a user must wait after calling "unlock"
+     * before they can withdraw their stake
+     */
+    uint256 public unlockDuration;
+
+    /**
+     * @notice Tracks funds that are in the process of being unlocked.
+     */
```
teinnt, line 36: Great =)))

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+    mapping(address => mapping(address => Unlocking)) public unlockings;
+
+    /** events **/
+    event UnlockDurationUpdated(uint256 unlockDuration);
+
+    /** errors **/
+    error SyloAddressCannotBeNil();
+    error NodeAddressCannotBeNil();
```
teinnt, line 44: Should be `ZeroAddress`, because the TS won't run the function if the address is undefined or null.

```suggestion
    error SyloTokenCannotBeZeroAddress();
    error NodeCannotBeZeroAddress();
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+    error UnlockDurationCannotBeZero();
+    error CannotStakeZeroAmount();
+    error CannotUnlockZeroAmount();
+    error CannotCancelUnlockingZeroAmount();
+    error CannotUnlockMoreThanStaked(uint256 stakeAmount, uint256 unlockAmount);
+    error CannotTransferMoreThanStaked(uint256 stakeAmount, uint256 transferAmount);
```
teinnt, line 50: Just to be clear:

```suggestion
    error CannotUnlockMoreThanStakedAmount(uint256 stakeAmount, uint256 unlockAmount);
    error CannotTransferMoreThanStakedAmount(uint256 stakeAmount, uint256 transferAmount);
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+    error StakeNotYetUnlocked();
+
+    function initialize(IERC20 sylo, uint256 _unlockDuration) external initializer {
```
teinnt, line 53:
```suggestion
    function initialize(IERC20 syloToken, uint256 _unlockDuration) external initializer {
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+    }
+
+    function _setUnlockDuration(uint256 _unlockDuration) internal {
+        if (_unlockDuration == 0) {
+            revert UnlockDurationCannotBeZero();
+        }
+
+        unlockDuration = _unlockDuration;
+        emit UnlockDurationUpdated(_unlockDuration);
+    }
```
teinnt, line 84: If we only called it twice here, I think we can make the function public so it is easier to maintain and reduce gas cost.
```suggestion
    function setUnlockDuration(uint256 _unlockDuration) public {
        if (_unlockDuration == 0) {
            revert UnlockDurationCannotBeZero();
        }

        unlockDuration = _unlockDuration;
        emit UnlockDurationUpdated(_unlockDuration);
    }
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+
+    /**
+     * @notice Called by stakers to add stake to a given node.
+     * This function will fail under the following conditions:
+     *   - If the Node address is invalid
+     *   - If the specified stake value is zero
```
teinnt, line 90:
```suggestion
     *   - If the stake amount is zero
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+    function _addStake(address node, uint256 amount) internal {
+        // update staking entry
+        StakeEntry storage stakeEntry = stakes[node].entries[msg.sender];
+        stakeEntry.amount += amount;
+        stakeEntry.updatedAt = block.timestamp;
+
+        // update total managed stake for this node
+        stakes[node].totalManagedStake += amount;
+
+
```
teinnt, line 120:
```suggestion
```

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+     * stake that was already in the unlocking phase will have the specified
+     * amount added to it, and its duration refreshed. This function will fail
+     * under the following conditions:
+     *   - If the Node address is invalid
+     *   - If no stake exists for the caller
+     *   - If the unlocking amount is zero
+     *   - If the unlocking amount is more than what is staked
+     * @param node The address of the node.
+     * @param amount The amount of stake to unlock in SOLO.
+     */
```
teinnt, line 134: Great comment!!!

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+        // update total stake managed by this contract
+        totalManagedStake -= amount;
+
+        // update unlocking
+        Unlocking storage unlocking = unlockings[node][msg.sender];
+
+        uint256 unlockAt = block.timestamp + unlockDuration;
+        if (unlocking.unlockAt < unlockAt) {
+            unlocking.unlockAt = unlockAt;
+        }
```
teinnt, line 166: May be we should have a comment here, saying 
"update `unlocking.unlockAt` in case the `unlockDuration` is updated"

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+     * @param node The address of the node.
+     */
+    function withdrawStake(address node) external {
+        if (node == address(0)) {
+            revert NodeAddressCannotBeNil();
+        }
+
+        Unlocking storage unlocking = unlockings[node][msg.sender];
+
+        if (unlocking.unlockAt >= block.timestamp) {
```
teinnt, line 185: Why we want to revert when `unlockAt == block.timestamp`? I think we should make sure other places use the same rule.

contracts/staking/sylo/SyloStakingManager.sol
```diff
@@ -0,0 +1,280 @@
...
+
+      stakeEntry.amount -= amount;
+      stakeEntry.updatedAt = block.timestamp;
+
+      _addStake(to, amount);
+    }
+
+    function getManagedStake(
+        address node,
+        address user
```
teinnt, line 264: Should we name it `staker` instead of `user`?

teinnt (commented):
Nothing wrong with the logic! Very awesome work Nick!

contracts/staking/seekers/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,202 @@
...
+    }
+
+    /**
+     * @notice Returns true if the contract implements the interface defined by
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return
+            interfaceId == type(ISeekerStatsOracle).interfaceId ||
+            super.supportsInterface(interfaceId);
```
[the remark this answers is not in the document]
Nick95550: This checks that the contract implements the ERC165 interface

contracts/staking/seekers/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,202 @@
...
+    /**
+     * @notice Sets the oracle account.
+     * @param _oracle The oracle account.
+     */
+    function setOracle(address _oracle) external onlyOwner {
+        if (_oracle == address(0)) {
+            revert OracleAddressCannotBeNil();
+        }
+        oracle = _oracle;
+    }
```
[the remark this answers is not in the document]
Nick95550: I didn't think it was needed in my opinion. Do you think it is required ?

contracts/staking/seekers/SeekerStatsOracle.sol
```diff
@@ -0,0 +1,202 @@
...
+        int256 totalChip = 0;
+
+        Seeker memory defaultSeeker;
+
+        for (uint256 i = 0; i < seekers.length; i++) {
+            Seeker memory seeker = seekers[i];
+            Seeker memory registeredSeeker = seekerStats[seeker.seekerId];
+            if (keccak256(abi.encode(registeredSeeker)) == keccak256(abi.encode(defaultSeeker))) {
+                revert SeekerNotRegistered(seeker.seekerId);
+            }
```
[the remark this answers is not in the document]
Nick95550: Sorry I don't quite understand, add a comment describing what the check is for ?

contracts/staking/seekers/SeekerStatsOracle.sol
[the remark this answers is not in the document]
JCSanPedro: Hmm actually I think we don't want to check if the parent contract implements the given interfaceId. I don't think there is much value in confirming if it conforms to the given `interfaceId`

contracts/staking/sylo/SyloStakingManager.sol
JCSanPedro: The reason I named it `Nil` is because that in Solidity, a zero-address does not always mean that no value was passed in or set. So I think using `Nil` makes it more clear that the method revert unless a non-null value is passed in.

Also, it is wrong to think in the context of TS when implementing Solidity. TS is just one way to write code to interact with Solidity.

In other languages, instead of passing in `ethers.ZeroAddress`, you could pass in empty `bytes`, and the solidity run time will decode that into the zero address (default value for null address).

contracts/staking/sylo/SyloStakingManager.sol
JCSanPedro: Hmm honestly seems unnecessarily verbose to me.

contracts/staking/sylo/SyloStakingManager.sol
JCSanPedro: I think `external` methods are cheaper than `public` methods https://ethereum.stackexchange.com/questions/19380/external-vs-public-best-practices

contracts/staking/sylo/ISyloStakingManager.sol
JCSanPedro: Fair point. I think we can keep that in mind in the future, and in general try to write code that is self documenting.

contracts/staking/sylo/SyloStakingManager.sol
JCSanPedro: hmm yes, seems like it should only revert when it is greater

contracts/staking/sylo/SyloStakingManager.sol
JCSanPedro: I think `node`/`user` is more clear, and easier to read than `stakee`/`staker`

contracts/staking/seekers/SeekerStatsOracle.sol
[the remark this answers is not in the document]
teinnt: Yeah, for me I think should be. Because we wanna keep it consistent whenever we update parameter.

contracts/staking/seekers/SeekerStatsOracle.sol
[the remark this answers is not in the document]
teinnt: That's okay. I mean a developer will need time to understand. Oh, why `SeekerNotRegistered` but use `==`? Then they will read the check carefully, and understand, oh, because compare with `defaultSeeker`! -> It takes time (actually I was like that). We can just add a comment above, like: `If registeredSeeker is defaultSeeker -> Seeker is not registered`

contracts/staking/sylo/SyloStakingManager.sol
teinnt: Gotcha

contracts/staking/sylo/SyloStakingManager.sol
teinnt: Great, I will keep that in mind for later consistency.

teinnt (approved):
Perfect!!!

PR #172, opened 2024-06-07
Sm 352 seeker staking

Nick95550:
Implements the interface and contract for Seeker Staking.


contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,227 @@
...
+        stakedSeekersByUser[msg.sender].push(seeker.seekerId);
+    }
+
+    /**
+     * @notice Transfers a staked seeker between nodes
+     * @param fromNode Address of node to transfer the staked seeker from
+     * @param toNode Address of node to transfer the staked seeker to
+     * @param seekerId Seeker ID of the staked seeker to transfer
+     */
+    function transferStakedSeeker(address fromNode, address toNode, uint256 seekerId) external {
```
JCSanPedro, line 120: I realise there is probably not a need for a transfer function, since there is no unlocking period for staked seekers 😓

contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,227 @@
...
+            oracle.registerSeeker(seeker, seekerStatsProof);
+        }
+
+        stakedSeekersById[seeker.seekerId] = StakedSeeker({
+            seekerId: seeker.seekerId,
+            node: node,
+            user: msg.sender
+        });
+        stakedSeekersByNode[node].push(seeker.seekerId);
+        stakedSeekersByUser[msg.sender].push(seeker.seekerId);
```
JCSanPedro, line 111: One thing we also need to take into account is what happens if the Seeker is already staked and then you stake it again. It looks like with the current implementation if you stake the same seeker twice, you end up with both being staked. However what should happen is that on the second stake call, it effectively unstakes the original stake.

I would recommend writing an internal `_unstake` method that unstakes the seekers (deletes the values in `stakedSeekersById`, `stakedSeekersByNode` and `stakedSeekersByUser`), and calling this `_unstake` method both when staking and unstaking

contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,227 @@
...
+            super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @notice Stakes a seeker
+     * @param node Address of node to stake seeker against
+     * @param seeker The object containing the seekers statistics
+     * @param seekerStatsProof The signature of the seekers proof message, signed by the oracle account.
+     */
+    function stakeSeeker(
```
JCSanPedro, line 87: I think because we want users to be able to stake multiple seekers in a single transaction, we need a method like this as well
```
stakeSeekers(address node, SeekerStatsOracle.Seeker[] calldata seekers, bytes[] calldata seekerStatProofs)
```
and you can move the implementation for `stakeSeeker` into an internal method `_stakeSeeker`

```
function stakeSeeker(node, seeker, seekerStatsProof) {
  _stakeSeeker(node, seeker, seekerStatsProof);
}

function stakeSeekers(node, seekers, seekerStatsProofs) {
  for (uint256 i = 0; seekers.length; ++i) {
    _stakeSeeker(node, seekers[i], seekerStatsProofs[i]);
  }
}
```

contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,227 @@
...
+            revert SeekerNotYetStaked();
+        }
+        if (stakedSeekersById[seekerId].user != msg.sender) {
+            revert SeekerNotStakedBySender();
+        }
+        if (stakedSeekersById[seekerId].node != fromNode) {
+            revert SeekerNotStakedToNode();
+        }
+
+        for (uint256 i = 0; i < stakedSeekersByNode[fromNode].length; i++) {
```
JCSanPedro, line 146: We should use `++i` instead of `i++`.https://ethereum.stackexchange.com/questions/133161/why-does-i-cost-less-gas-than-i

JCSanPedro (changes requested):
Nice work 👍

contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,242 @@
...
+            if (seekerStatsProof.length == 0) {
+                revert SeekerProofIsEmpty();
+            }
+            oracle.registerSeeker(seeker, seekerStatsProof);
+        }
+
+        StakedErrors err = _isSeekerStakedError(node, seeker.seekerId);
+        if (err == StakedErrors.NIL) {
+            _unstakeSeeker(node, seeker.seekerId);
+        }
```
teinnt, line 139: Hmm, I dont think we should auto unstake for them. They should do this themselves, or just transfer the seeker ownership. This code needs a comment to warn users, here or in the `@notice` comment. But I think we should just revert here instead of auto unstake for the user. It makes the code clearer and be strict for what it should do. What do you think?

contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,242 @@
...
+
+        _unstakeSeeker(node, seekerId);
+    }
+
+    /**
+     * @notice Unstake a seeker
+     * @param node Address of node to unstake seeker from
+     * @param seekerId Seeker ID of staked seeker
+     */
```
teinnt, line 181: Similar to `_stakeSeeker`, should we remove comment here too? Or we should add more info, like: `replaces the last stakedSeeker with the unstakedSeeker index in stakedSeekersByNode and stakedSeekersByUser arrays and delete it in the stakedSeekersById`

Up to you.

BTW GREAT FUNCTION!!!

contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,242 @@
...
+        }
+        if (!oracle.isSeekerRegistered(seeker)) {
+            if (seekerStatsProof.length == 0) {
+                revert SeekerProofIsEmpty();
+            }
+            oracle.registerSeeker(seeker, seekerStatsProof);
+        }
+
+        StakedErrors err = _isSeekerStakedError(node, seeker.seekerId);
+        if (err == StakedErrors.NIL) {
```
teinnt, line 137:
```suggestion
        if (_isSeekerStakedError(node, seeker.seekerId) == StakedErrors.NIL) {
```

contracts/staking/seekers/SeekerStakingManager.sol
```diff
@@ -0,0 +1,242 @@
...
+            return StakedErrors.SEEKER_NOT_STAKED_BY_SENDER;
+        }
+        if (stakedSeekersById[seekerId].node != node) {
+            return StakedErrors.SEEKER_NOT_STAKED_TO_NODE;
+        }
+        return StakedErrors.NIL;
+    }
+
+    /**
+     * @notice Get a staked seeker by node address
```
teinnt, line 228:
```suggestion
     * @notice Get all seekers that were staked to a specific node address
```

teinnt (commented):
Wow awesome!!! Very cool PR!!! I am glad to review this PR! Thanks Nick!!!!!

contracts/staking/seekers/SeekerStakingManager.sol
Nick95550: I actually did this on Johns recommendation, I think I tend to agree with your pov though. https://github.com/futureversecom/sylo-ethereum-contracts/pull/172#discussion_r1632416832

contracts/staking/seekers/SeekerStatsOracle.sol
```diff
@@ -116,13 +126,9 @@ contract SeekerStatsOracle is ISeekerStatsOracle, Initializable, Ownable2StepUpg
...
     /**
-     * @notice Creates a proofing message unique to the provided seeker.
+     * @notice Registers a seeker only callable from oracle
      * @param seeker The object containing the seekers statistics.
      */
-    function createProofMessage(Seeker calldata seeker) external pure returns (bytes memory) {
-        return _createProofMessage(seeker);
-    }
-
     function registerSeekerRestricted(Seeker calldata seeker) external {
```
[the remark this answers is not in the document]
Nick95550: hmm I think that restricted here applies to the oracle account but the setOracle is restrictred to the deployer, so they are different ?

contracts/staking/seekers/SeekerStakingManager.sol
JCSanPedro: I don't think the user should have to `unstake` then `stake` again (2 transactions) just to move their staked seeker.

I agree that the @notice comment should mention that it will unstake the seeker if it is already staked.

teinnt (approved):
LGTM!!!! GREAT PR!!!!

PR #173, opened 2024-06-12
Sm 360 implement protocol time manager

Nick95550:
Implements the interface and contract for protocol time manager.


contracts/ProtocolTimeManager.sol
```diff
@@ -0,0 +1,251 @@
...
+     * @notice Sets the start time for the cycle/period intervals
+     * @param _start The start time for the cycle/period intervals in unix
+     */
+    function setProtocolStart(uint256 _start) external onlyOwner {
+        if (_start == 0) {
+            revert CannotSetProtocolStartWithZeroStart();
+        }
+        if (_start > block.timestamp) {
+            revert CannotSetStartInFuture();
+        }
```
JCSanPedro, line 88: We would actually want to be able to set it for a future block. What we don't want is to be able to set it once the protocol has already started

contracts/ProtocolTimeManager.sol
```diff
@@ -0,0 +1,251 @@
...
+     * @notice Get the start time of the cycle/period intervals
+     */
+    function getStart() external view returns (uint256) {
+        return start;
+    }
+
+    /**
+     * @notice Get the cycle and period durations
+     */
+    function timeNow() external view returns (uint256, uint256) {
```
JCSanPedro, line 246:
```suggestion
    function now() external view returns (uint256, uint256) {
```

Any reason we can't just name it `now`?

Nick95550: although now() was deprecated it seems that the keyword is still in use as warning

PR #174, opened 2024-06-13
Sm 361 reintroduce registries

Nick95550:
This PR reintroduces the registries contract and its test cases into the new contract suite. Also adds a test case for supporting erc165 to the contracts.

### DIFF
--- a/contracts/Registries.sol
+++ b/contracts/Registries.sol
@@ -1,0 +1,177 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/utils/Strings.sol";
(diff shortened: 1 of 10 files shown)


contracts/Registries.sol
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * @notice Helper function for deriving the proof message used to
+     * validate seeker ownership.
+     * @param seekerId The tokenId of the seeker used for operation.
+     * @param node The address of the node which that will be operated
+     * by the specified seeker.
+     * @param nonce The nonce used for this message.
+     */
+    function getProofMessage(
```
JCSanPedro, line 162: I think this method is no longer needed

JCSanPedro (approved):

PR #176, opened 2024-06-17
Sm 363 partially implementation rewards manager

Nick95550:
Partially implement the rewards manager. 

NOTE: this PR also adds the ticketing contract to test access control for the rewards manager contract

### DIFF
--- a/contracts/RewardsManager.sol
+++ b/contracts/RewardsManager.sol
@@ -1,0 +1,157 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts/access/AccessControl.sol";
+
+import "./IRewardsManager.sol";
+import "./Registries.sol";
+import "./Ticketing.sol";
+import "./ITicketing.sol";
+
+contract RewardsManager is IRewardsManager, Initializable, AccessControl {
+    /**
+     * @notice The only ticketing role given to the ticketing contract
+     * to restrict access to the incrementRewardPool function
+     */
+    bytes32 public constant onlyTicketing = keccak256("ONLY_TICKETING");
+
+    /**
+     * @notice Registries contract
+     */
+    Registries public registries;
+
+    /**
+     * @notice Ticketing contract
+     */
+    Ticketing public ticketing;
+
+    /**
+     * @notice Tracks claims from staker accounts
+     */
+    mapping(address => mapping(address => uint256)) claims;
+
+    /**
+     * @notice Tracks reward pools for each reward cycle
+     */
+    mapping(address => mapping(uint256 => uint256)) rewardPools;
+
+    /**
+     * @notice Tracks the nodes unclaimed commission on redeemed tickets
+     */
+    mapping(address => uint256) unclaimedNodeCommission;
+
+    error CannotInitializeEmptyRegistriesAddress();
+    error CannotInitializeEmptyTicketingAddress();
+    error CannotIncrementRewardPoolWithZeroNodeAddress();
+    error CannotIncrementRewardPoolWithZeroAmount();
+    error CannotInitializeWithNonTicketing();
+
+    function initialize(Registries _registries, Ticketing _ticketing) external initializer {
+        if (address(_registries) == address(0)) {
+            revert CannotInitializeEmptyRegistriesAddress();
+        }
+        if (address(_ticketing) == address(0)) {
+            revert CannotInitializeEmptyTicketingAddress();
+        }
+        if (!ERC165(address(_ticketing)).supportsInterface(type(ITicketing).interfaceId)) {
+            revert CannotInitializeWithNonTicketing();
+        }
+
+        registries = _registries;
+        ticketing = _ticketing;
+
+        _grantRole(onlyTicketing, address(_ticketing));
+    }
+
+    /**
+     * @notice Returns true if the contract implements the interface defined by
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return
+            interfaceId == type(IRewardsManager).interfaceId ||
+            super.supportsInterface(interfaceId);
+    }
+
+    /**
+     * @notice Increments a nodes reward pool. Callable only by the ticketing
+     * contract when redeeming a ticket.
+     * @param node Address of the node
+     * @param cycle Current reward cycle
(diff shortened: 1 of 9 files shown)


contracts/RewardsManager.sol
```diff
@@ -0,0 +1,157 @@
...
+     * @notice Tracks reward pools for each reward cycle
+     */
+    mapping(address => mapping(uint256 => uint256)) rewardPools;
+
+    /**
+     * @notice Tracks the nodes unclaimed commission on redeemed tickets
+     */
+    mapping(address => uint256) unclaimedNodeCommission;
+
+    error CannotInitializeEmptyRegistriesAddress();
```
JCSanPedro, line 44:
```suggestion
    error RegistriesAddressCannotBeNil();
```

JCSanPedro: We should try to be consistent with the naming conventions. I think it differs in other places as well, but let's fix it in this contract at least

contracts/RewardsManager.sol
```diff
@@ -0,0 +1,157 @@
...
+        uint256 amount
+    ) external onlyRole(onlyTicketing) {
+        if (address(0) == node) {
+            revert CannotIncrementRewardPoolWithZeroNodeAddress();
+        }
+        if (amount == 0) {
+            revert CannotIncrementRewardPoolWithZeroAmount();
+        }
+
+        if (registries.defaultPayoutPercentage() != 0) {
```
JCSanPedro, line 96: I think it is is valid for the value to be `0` (100% of rewards go to node).

contracts/RewardsManager.sol
```diff
@@ -0,0 +1,157 @@
...
+
+    /**
+     * @notice Multiply a value by a given percentage. Converts the provided
+     * uint128 value to uint256 to avoid any reverts on overflow.
+     * @param value The value to multiply.
+     * @param percentage The percentage, as a ratio of 100000.
+     */
+    function percOf(uint256 value, uint32 percentage) internal pure returns (uint256) {
+        return (uint256(value) * percentage) / 100000;
+    }
```
JCSanPedro, line 156: We should add back the `SyloUtils` library to the repo

contracts/RewardsManager.sol
JCSanPedro: I think the `percOf` function should be able to support 0%, so shouldn't need this `if` statement

contracts/RewardsManager.sol
```diff
@@ -0,0 +1,157 @@
...
+     * @notice Gets the reward pool for a node from a specific cycle
+     * @param node Address of nodes
+     * @param cycle Associated reward pools cycle
+     */
+    function getRewardPool(address node, uint256 cycle) external view returns (uint256) {
+        return rewardPools[node][cycle];
+    }
+
+    /**
+     * @notice Gets the reward pool for a node from a specific cycle
```
JCSanPedro, line 116: I think this comment is slightly wrong

contracts/Ticketing.sol
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * @notice Returns true if the contract implements the interface defined by
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(ITicketing).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    function testerIncrementRewardPool(address node, uint256 cycle, uint256 amount) external {
```
JCSanPedro, line 34: I think for testing purposes, it would be more clean to just call `grantRole` https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol#L122 to the deployer account. Then in your unit tests you can just call `incrementRewardPool` directly

```
beforeEach(asyc () => {
  await rewardsManager.grantRole(accounts[0].getAddress());
})

it("...", async () => {
  await rewardsManager.incrementRewardPool(...)
})
```

JCSanPedro: The approach is better because it makes the unit tests more "unit" like. Reading the test, you can clearly see the function being tested (instead of assuming it is being called correctly by the `testerIncrementRewardPool` method)

test/rewardsManager.test.ts
```diff
@@ -0,0 +1,347 @@
...
+
+    await ticketing.testerIncrementRewardPool(await node1.getAddress(), 0, 100);
+    await ticketing.testerIncrementRewardPool(await node1.getAddress(), 1, 200);
+    await ticketing.testerIncrementRewardPool(await node2.getAddress(), 0, 300);
+    await ticketing.testerIncrementRewardPool(await node2.getAddress(), 1, 500);
+
+    // const rewardPoolNode1Cycle1 = await rewardsManager.getRewardPool(
+    //   await node1.getAddress(),
+    //   0,
+    // );
```
JCSanPedro, line 181: Lets clean up these comments

test/rewardsManager.test.ts
```diff
@@ -0,0 +1,347 @@
...
+    );
+
+    assert.equal(
+      invalid,
+      false,
+      'Expected rewards manager to not support incorrect interface',
+    );
+  });
+
+  async function checkInitialRewardPoolState(_rewardsManager: RewardsManager) {
```
JCSanPedro, line 310: Honestly I don't see the need for this function. Like what value do you add (i.e what additional functionality or cases are you testing) by calling this in more than 1 test?

JCSanPedro: You probably want something more similar to what we do in the staking manager unit tests https://github.com/futureversecom/sylo-ethereum-contracts/blob/v2/seeker-staking/test/staking/syloStakingManager.test.ts#L642.

Where you validate an expected initial reward pool state, then you do something, then have an expected post reward pool state.

test/rewardsManager.test.ts
```diff
@@ -0,0 +1,347 @@
...
+      'RewardsManager',
+    );
+    const registriesFactory = await ethers.getContractFactory('Registries');
+    const ticketingFactory = await ethers.getContractFactory('Ticketing');
+
+    const newRewardsManager = await rewardsManagerFactory.deploy();
+    const newRegistries = await registriesFactory.deploy();
+    const newTicketing = await ticketingFactory.deploy();
+
+    await newRegistries.initialize(0);
```
JCSanPedro, line 338: I already fixed this in the branch for the ticketing contract, but no need to implement this `initialiseContracts` method just so you can have a new registries contract with the payout rate set to 0 (just call `setDefaultPayoutPercentage(0)` at the start of the tests that need it)

JCSanPedro (approved):

PR #177, opened 2024-06-26
Sm 368 implement directories contract

Nick95550:
Implement the directories interface

### DIFF
--- a/contracts/Directory.sol
+++ b/contracts/Directory.sol
@@ -1,0 +1,148 @@
+// SPDX-License-Identifier: Apache-2.0
+pragma solidity ^0.8.18;
+
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "@openzeppelin/contracts-upgradeable/access/Ownable2StepUpgradeable.sol";
+import "@openzeppelin/contracts/utils/introspection/ERC165.sol";
+import "./staking/StakingOrchestrator.sol";
+import "./ProtocolTimeManager.sol";
+
+import "./IProtocolTimeManager.sol";
+import "./IDirectory.sol";
+
+contract Directory is IDirectory, Initializable, Ownable2StepUpgradeable, ERC165 {
+    StakingOrchestrator public stakingOrchestrator;
+    ProtocolTimeManager public protocolTimeManager;
+
+    /**
+     * @notice Tracks each directory, these directories are apart of
+     * each staking period for each reward cycle
+     */
+    mapping(uint256 => mapping(uint256 => Directory)) public directories;
+
+    error CannotInitialiseWithZeroStakingOrchestratorAddress();
+    error CannotInitialiseWithZeroProtocolTimeManagerAddress();
+    error CannotJoinDirectoryWithZeroStake();
+    error StakeeAlreadyJoinedDirectory();
+
+    function initialize(
+        StakingOrchestrator _stakingOrchestrator,
+        ProtocolTimeManager _protocolTimeManager
+    ) external initializer {
+        Ownable2StepUpgradeable.__Ownable2Step_init();
+
+        if (address(_stakingOrchestrator) == address(0)) {
+            revert CannotInitialiseWithZeroStakingOrchestratorAddress();
+        }
+        if (address(_protocolTimeManager) == address(0)) {
+            revert CannotInitialiseWithZeroProtocolTimeManagerAddress();
+        }
+
+        stakingOrchestrator = _stakingOrchestrator;
+        protocolTimeManager = _protocolTimeManager;
+    }
+
+    /**
+     * @notice Returns true if the contract implements the interface defined by
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(IDirectory).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    function scan(uint128 point) external view returns (address) {
+        uint256 currentStakingPeriod = protocolTimeManager.getCurrentPeriod();
+        IProtocolTimeManager.Cycle memory cycle = protocolTimeManager.getCurrentCycle();
+        return _scan(point, cycle.iteration, currentStakingPeriod);
+    }
+
+    function scanWithTime(
+        uint128 point,
+        uint256 rewardCycleId,
+        uint256 stakingPeriodId
+    ) external view returns (address) {
+        return _scan(point, rewardCycleId, stakingPeriodId);
+    }
+
+    /**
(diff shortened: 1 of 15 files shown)


contracts/Directory.sol
```diff
@@ -0,0 +1,148 @@
...
+     * @notice Returns true if the contract implements the interface defined by
+     * `interfaceId` from ERC165.
+     */
+    function supportsInterface(bytes4 interfaceId) public view virtual override returns (bool) {
+        return interfaceId == type(IDirectory).interfaceId || super.supportsInterface(interfaceId);
+    }
+
+    function scan(uint128 point) external view returns (address) {
+        uint256 currentStakingPeriod = protocolTimeManager.getCurrentPeriod();
+        IProtocolTimeManager.Cycle memory cycle = protocolTimeManager.getCurrentCycle();
```
JCSanPedro, line 55: Can replace these with just `protocolTimeManager.getTime()`?

contracts/Directory.sol
```diff
@@ -0,0 +1,148 @@
...
+                // expectedVal >= upper
+                left = index + 1;
+            }
+        }
+    }
+
+    function joinNextDirectory() external {
+        IProtocolTimeManager.Cycle memory currentRewardCycle = protocolTimeManager
+            .getCurrentCycle();
+        uint256 currentStakingPeriod = protocolTimeManager.getCurrentPeriod();
```
JCSanPedro, line 116: Replace these with `getTime` as well

contracts/Directory.sol
```diff
@@ -0,0 +1,148 @@
...
+
+        uint256 nodeStake = stakingOrchestrator.getNodeCurrentStake(msg.sender);
+        if (nodeStake == 0) {
+            revert CannotJoinDirectoryWithZeroStake();
+        }
+
+        if (
+            directories[currentRewardCycle.iteration][currentStakingPeriod + 1].stakes[
+                msg.sender
+            ] > 0
```
JCSanPedro, line 126: The formatting looks strange here. Can you try run `npx prettier --write 'contracts/**/*.sol' `?

contracts/Directory.sol
```diff
@@ -0,0 +1,148 @@
...
+        ) {
+            revert StakeeAlreadyJoinedDirectory();
```
JCSanPedro, line 128:
```suggestion
            revert NodeAlreadyJoinedDirectory();
```

contracts/ProtocolTimeManager.sol
```diff
@@ -417,4 +417,14 @@ contract ProtocolTimeManager is
         uint256 period = _getCurrentPeriod();
         return (cycle.iteration, period, cycle);
     }
+
+    function isFinalStakingPeriod() external view returns (bool) {
```
JCSanPedro, line 421: I always feel it is odd whenever we need to add such a specific function to the interface. Like this method would only be called by the Directory contract for a very specific use case.

In these cases I would prefer to think of a more generic utility function that can fulfil more use cases. Though I think what we could do here is extend the `getTime` return value to also return the period duration. So we return
```
getTime returns (
  uint256 cycleIteration,
  uint256 period,
  Cycle cycle,
  uint256 periodDuration
)
```

Then that is enough information for any caller to determine if it is the final period or not.

test/authorizedAccount.test.ts
```diff
@@ -663,32 +661,35 @@ describe('Authorized Accounts', () => {
...
-      1000,
-    );
-    assert.equal(validate, true);
+    // validate = await authAccountsConnectMain.validatePermission(
+    //   mainAccountAddress,
+    //   delegatedAccount1,
+    //   Permission.PersonalSign,
+    //   1000,
+    // );
+    // assert.equal(validate, true);
```
JCSanPedro, line 685: Make sure to remove any commented code out

test/Directory.test.ts
```diff
@@ -0,0 +1,640 @@
...
+      if (!b) throw new Error('block undefined');
+      return b;
+    });
+
+    await protocolTimeManager.setProtocolStart(block.timestamp + time);
+
+    return protocolTimeManager.getStart().then(Number);
+  }
+
+  async function startProtocol() {
```
JCSanPedro, line 632: You should move this function to a utils file instead of duplicating it

test/Directory.test.ts
```diff
@@ -0,0 +1,640 @@
...
+    address = await directory.scan(0);
+    assert.equal(address, await nodeOne.getAddress());
+
+    await setTimeSinceStart(500);
+
+    address = await directory.scan(0);
+    assert.equal(address, ethers.ZeroAddress);
+  });
+
+  it('node joins next reward cycle', async () => {
```
JCSanPedro, line 127:
```suggestion
  it('node joins next cycle at the final period', async () => {
```

test/Directory.test.ts
```diff
@@ -0,0 +1,640 @@
...
+  it('should be able to scan for different staking periods', async () => {
+    let address: string;
+    const { setTimeSinceStart } = await startProtocol();
+    await stakingOrchestator.syloStakeAdded(nodeOne, ethers.ZeroAddress, 1000);
+    await directory.connect(nodeOne).joinNextDirectory();
+
+    await setTimeSinceStart(150);
+
+    address = await directory.scan(0);
+    assert.equal(address, await nodeOne.getAddress());
```
JCSanPedro, line 119: Feels like this pattern can be put into a function
```
const testScan = async function(point: number, expectedAddress: string) {
  expect(await directory.scan(point)).to.equal(expectedAddress);
}
```

test/Directory.test.ts
```diff
@@ -0,0 +1,640 @@
...
+    address = await directory.scan(0);
+    assert.equal(address, ethers.ZeroAddress);
+
+    await setTimeSinceStart(1950);
+    await directory.connect(nodeOne).joinNextDirectory();
+
+    await setTimeSinceStart(2050);
+
+    address = await directory.scan(0);
+    assert.equal(address, await nodeOne.getAddress());
```
JCSanPedro, line 158: The readability of this test could be improved a lot with some comments I think.

test/Directory.test.ts
```diff
@@ -0,0 +1,640 @@
...
+  });
+
+  it('should be able to scan for same staking period', async () => {
```
JCSanPedro, line 161:
```suggestion
  it('should be able to scan for previous periods', async () => {
```

This seems like what the test is actually checking for

JCSanPedro: But I would suggest using multiple nodes, and make sure each period has a different set of nodes join (so the directories are different for each period). Reason being that if you only had 1 node joining, and your `scanWithTime` implementation was wrong in the sense that it always scanned the current period instead of the specified time, then this test would pass.

JCSanPedro (approved):

PR #178, opened 2024-07-01
Sm 362 ticketing

Nick95550:
Reintroduce the ticketing contract


test/paymets/ticketing.test.ts
```diff
@@ -1,38 +1,57 @@
...
     contracts = await deployContracts();
     deposits = contracts.deposits;
     ticketing = contracts.ticketing;
+    authorizedAccounts = contracts.authorizedAccounts;
+    futurepass = contracts.futurepassRegistrar;
 
-    await deposits.approveTicketing(ticketing.getAddress());
+    await ticketing.setBaseLiveWinProb(2n ** 128n - 1n);
+    sufficientEscrow = (await ticketing.faceValue()) + BigInt(100);
+    insufficientEscrow = (await ticketing.faceValue()) - BigInt(100);
```
JCSanPedro, line 40: You can replace the `BigInt(100)` constructor with just `100n`

test/paymets/ticketing.test.ts
```diff
@@ -1,38 +1,57 @@
...
-      ticketing.initialize(
+      ticketingTemp.initialize(
         ethers.ZeroAddress,
         ethers.ZeroAddress,
         ethers.ZeroAddress,
         ethers.ZeroAddress,
         ethers.ZeroAddress,
         ethers.ZeroAddress,
-        100n,
+        insufficientEscrow,
```
JCSanPedro, line 56: Feels confusing to use `insufficientEscrow` here, as that value doesn't actually matter. Would just use `1n` to make it a bit more clear that the value is not significant.

test/paymets/ticketing.test.ts
```diff
@@ -99,7 +115,1670 @@ describe.only('Ticketing', () => {
...
+    ) => {
+      await expect(setterNonOwner(value)).to.revertedWith(
+        'Ownable: caller is not the owner',
+      );
+    };
+
+    await updateParam(ticketing.connect(accounts[1]).setFaceValue, 0n);
+
+    await updateParam(
+      ticketing.connect(accounts[1]).setMultiReceiverFaceValue,
```
JCSanPedro, line 157: Should be `setFaceValue`?

test/paymets/ticketing.test.ts
```diff
@@ -99,7 +115,1670 @@ describe.only('Ticketing', () => {
...
+    const senderSignature = await alice.signMessage(
+      ethers.getBytes(ticketHash),
+    );
+
+    const senderSig = createUserSignature(
+      signatureTypes.main,
+      senderSignature,
+      ethers.ZeroAddress,
+      createEmptyAttachedAuthorizedAccount(),
+    );
```
JCSanPedro, line 633: I think would be worth to create a util function `const { ticket, ticketHash, senderSig, receiverSig } = createSignedTicket(1, alice, bob, redeemer.address, block.number, redeemerRand)`

test/paymets/ticketing.test.ts
```diff
@@ -99,7 +115,1670 @@ describe.only('Ticketing', () => {
...
+        block.number,
+        await ticketing.faceValue(),
+      );
+
+    const depositAfter = await deposits.getDeposit(alice.address);
+    assert.equal(
+      depositAfter[0],
+      sufficientEscrow - (await ticketing.faceValue()),
+    );
+    assert.equal(depositAfter[1], sufficientEscrow);
```
JCSanPedro: A `redeemAndValidateDeposit` util method that contains this functionality would be good as well

JCSanPedro (changes requested):
Looking good. Needs a rebase. Just try to spend some time thinking about ways to reduce code duplication in the unit tests.

JCSanPedro (approved):

PR #183, opened 2024-07-31
SM-378 Auto-Join

JCSanPedro:
Mostly contains changes to the contract interfaces to make some queries simpler.

### DIFF
--- a/contracts/IDirectory.sol
+++ b/contracts/IDirectory.sol
@@ -29,9 +29,11 @@ interface IDirectory {
 
     function scanWithTime(
         uint128 point,
-        uint256 rewardCycleId,
-        uint256 stakingPeriodId
+        uint256 cycleId,
+        uint256 periodId
     ) external returns (address);
 
     function joinNextDirectory() external;
+
+    function getDirectoryStake(uint256 cycleId, uint256 periodId, address node) external view returns (uint256);
 }
--- a/contracts/IDirectory.sol
+++ b/contracts/IDirectory.sol
@@ -14,7 +14,7 @@ interface IDirectory {
     }
 
     /**
-     * @dev An EpochDirectory will be stored for every period. The
+     * @dev A Directory will be stored for every period. The
      * directory will be constructed piece by piece as Nodes join,
      * each adding their own directory entry based on their current
      * stake value.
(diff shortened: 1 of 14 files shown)


contracts/ProtocolTimeManager.sol
```diff
@@ -320,8 +320,10 @@ contract ProtocolTimeManager is
         uint256 totalTimeElapsedWithinPeriod = block.timestamp - cycle.start;
 
         uint256 periodDuration = _getPeriodDuration();
+        uint256 periodId = totalTimeElapsedWithinPeriod / periodDuration;
+        uint256 periodStart = cycle.start + periodDuration * periodId;
```
Nick95550, line 324: I think would be better described as 'totalTimeElapsedWithinCycle'

contracts/IDirectory.sol
```diff
@@ -29,9 +29,11 @@ interface IDirectory {
...
-        uint256 rewardCycleId,
-        uint256 stakingPeriodId
+        uint256 cycleId,
+        uint256 periodId
     ) external returns (address);
 
     function joinNextDirectory() external;
+
+    function getDirectoryStake(uint256 cycleId, uint256 periodId, address node) external view returns (uint256);
 }
```
Nick95550, line 39: We should change the description on the Directory struct to no longer reference epoch

Nick95550 (approved):
Nothing big just a couple minor comments

contracts/ProtocolTimeManager.sol
JCSanPedro: `totalTimeElapsedWithinCycle` I don't think that accurately reflects what the above value means.

For example, the current time might be `450` seconds into the cycle, but if the period duration is `100s` and we are in the 4th period, then the start of the period is actually `400`.

contracts/ProtocolTimeManager.sol
```diff
@@ -320,8 +320,10 @@ contract ProtocolTimeManager is
         uint256 totalTimeElapsedWithinPeriod = block.timestamp - cycle.start;
 
         uint256 periodDuration = _getPeriodDuration();
+        uint256 periodId = totalTimeElapsedWithinPeriod / periodDuration;
```
JCSanPedro, line 323:
```suggestion
        uint256 periodId = totalTimeElapsedWithinCycle / periodDuration;
```
blah blah blah

PR #185, opened 2024-08-05
SM-386

JCSanPedro:
This is a small PR that adds the following methods to the `RewardsManager` contract.

- `getClaim`
- `getUnclaimedReward`
- `getUnclaimedReward`

These methods make it easier for clients to determine if a user has an unclaimed reward for a given cycle.

### DIFF
--- a/contracts/payments/RewardsManager.sol
+++ b/contracts/payments/RewardsManager.sol
@@ -132,6 +132,28 @@ contract RewardsManager is IRewardsManager, Initializable, AccessControl {
         return rewardPools[node][cycle];
     }
 
+    function getUnclaimedReward(address node, address user, uint256 cycle) external view returns (uint256) {
+        return _getUnclaimedReward(node, user, cycle);
+    }
+
+    function _getUnclaimedReward(address node, address user, uint256 cycle) internal view returns (uint256) {
+        if (claims[node][user][cycle]) {
+            return 0;
+        }
+
+        return _getClaim(node, user, cycle);
+    }
+
+    function getUnclaimedRewards(address node, address user, uint256[] calldata cycles) external view returns (uint256[] memory) {
+        uint256 [] memory unclaimedRewards = new uint256[](cycles.length);
+
+        for (uint256 i = 0; i < cycles.length; i++) {
+            unclaimedRewards[i] = _getUnclaimedReward(node, user, cycles[i]);
+        }
+
+        return unclaimedRewards;
+    }
+
     function getClaim(address node, address user, uint256 cycle) external view returns (uint256) {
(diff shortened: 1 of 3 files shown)


contracts/payments/RewardsManager.sol
```diff
@@ -132,6 +132,28 @@ contract RewardsManager is IRewardsManager, Initializable, AccessControl {
...
+
+    function getUnclaimedRewards(address node, address user, uint256[] calldata cycles) external view returns (uint256[] memory) {
+        uint256 [] memory unclaimedRewards = new uint256[](cycles.length);
+
+        for (uint256 i = 0; i < cycles.length; i++) {
+            unclaimedRewards[i] = _getUnclaimedReward(node, user, cycles[i]);
+        }
+
+        return unclaimedRewards;
+    }
```
Nick95550, line 155: Nit:
```suggestion

    function getUnclaimedReward(address node, address user, uint256 cycle) external view returns (uint256) {
        return _getUnclaimedReward(node, user, cycle);
    }

    function getUnclaimedRewards(address node, address user, uint256[] calldata cycles) external view returns (uint256[] memory) {
        uint256 [] memory unclaimedRewards = new uint256[](cycles.length);

        for (uint256 i = 0; i < cycles.length; i++) {
            unclaimedRewards[i] = _getUnclaimedReward(node, user, cycles[i]);
        }

        return unclaimedRewards;
    }

    function _getUnclaimedReward(address node, address user, uint256 cycle) internal view returns (uint256) {
        if (claims[node][user][cycle]) {
            return 0;
        }

        return _getClaim(node, user, cycle);
    }
```

test/paymets/rewardsManager.test.ts
```diff
@@ -308,6 +319,16 @@ describe('Rewards Manager', () => {
...
       }
 
+      const unclaimedRewards = await rewardsManager.getUnclaimedRewards(
+        node1.getAddress(),
+        user.getAddress(),
+        cycles,
+      );
+
+      for (const [i, r] of unclaimedRewards.entries()) {
+        expect(r).to.be.eq(BigInt(i + 1) * rewardAmount);
```
Nick95550, line 329: Nit:

```suggestion
        expect(r).to.be.equal(BigInt(i + 1) * rewardAmount);
```

Nick95550 (approved):
Looking good only minors

test/paymets/rewardsManager.test.ts
JCSanPedro: I think `.eq` is not ambiguous at all

Nick95550 (approved):
LGTM

PR #186, opened 2024-08-11
SM-385 CLI

JCSanPedro:
This is a supporting PR for SM-385. It's main contribution is that it re-implements the deployment script, and also the initialisation script for the local test network.

We also move from `ganache -> foundry anvil` for the local test node. Ganache has been deprecated, whilst anvil is still being supported, and allows us to save chain state with the `--state state.json` option.

Testing the CLI has also prompted some minor changes/fixes in the contract code, which has been included in this PR as well. These changes include:

* Allowing protocol time manager to get next before protocol has started
* Adding method to Directory to get all entries for a given directory


contracts/staking/StakingOrchestrator.sol
```diff
@@ -224,6 +230,7 @@ contract StakingOrchestrator is IStakingOrchestrator, Initializable, AccessContr
         _grantRole(DEFAULT_ADMIN_ROLE, msg.sender);
         _grantRole(onlyOwner, msg.sender);
         _grantRole(onlyStakingManager, address(_syloStakingManager));
+        _grantRole(onlyStakingManager, address(_seekerStakingManager));
```
Nick95550, line 233: I think we should create a seperate role for this

contracts/Directory.sol
```diff
@@ -141,4 +136,22 @@ contract Directory is IDirectory, Initializable, Ownable2StepUpgradeable, ERC165
...
+        DirectoryEntry memory entry;
+        DirectoryEntry[] memory entries = directories[cycleId][periodId].entries;
+
+        for (uint256 i; i < entryLength; ++i) {
+            entry = entries[i];
+            nodes[i] = entry.stakee;
+            boundaries[i] = entry.boundary;
+        }
+
+        return (nodes, boundaries);
```
Nick95550, line 155: Whats the reason for storing a varibale in entry here ?

docker/localhost-with-protocol/README.md
```diff
@@ -0,0 +1,58 @@
...
+yarn docker-ganache
+```
+
+## Run docker
+
+The options used to run the local ganache node should be set correctly
+to ensure that it re-uses the same directory for the chain state.
+
+```sh
+docker run -p 8545:8545 dn3010/sylo-ethereum-testnet:0.1.0 `bash -c ganache-cli --host 0.0.0.0 --db ganache-data --blockTime 5`
```
Nick95550, line 57: I think these need editing

contracts/Directory.sol
Nick95550: Also what is the benefit in returning the boundaries ?

Nick95550 (changes requested):
Just a couple minors

contracts/staking/StakingOrchestrator.sol
JCSanPedro: It's intended for the `onlyStakingManager` to cover both roles. Seems redundant to define two distinct roles

contracts/Directory.sol
JCSanPedro: It is just a generally useful utility method for debugging. Allows us to see all of the nodes that have joined a directory, and what staking value they joined with.

JCSanPedro: It also allows a local client to perform the scan method locally which is benefecial if it knows it will need to call scan multiple times for the same directory.

